Issue #5, opened 2014-09-18
Enterprise-specific Values


parkr:
Hey @benbalter! 

So for things like `hostname`, how would you like this program to fetch those? `ENV` vars? Hypothetically we can write in `enterprise? ? 'ghe.io' : 'github.com'` but that doesn't give very much flexibility.

The variables that are custom based on the env are:

- `hostname`
- `pages_hostname`
- `api_url`
- `environment`

How do you want this public code to handle that stuff?

parkr:
Falling back to Octokit configuration for this kind of thing.

PR #10, opened 2015-06-15
Allow pages values to be overwritable

parkr:
Support GitHub Enterprise.

/cc #5 @benbalter


parkr:
@benbalter any thoughts?

benbalter:
What do you think about using/mirroring [the Octokit env vars](https://github.com/octokit/octokit.rb#using-env-variables)?

parkr:
> What do you think about using/mirroring the Octokit env vars?

A much easier route! Think they'll always be the same? If so, we can just delegate to Octokit.

parkr:
@benbalter Updated ^^

PR #17, opened 2015-09-03
Shut off auto_paginate

### LINKED ISSUE #16: Public Repos (and other collections) only have first 30 results?
In my markdown files, I am trying to display the names of all public repositories for a particular organization, but it appears that the collection values available in `site.github` only contain 30 entries (likely based on the API's spec that says it [by default] only returns 30).

    {% for repository in site.github.public_repositories %}
            <li>{{ repository.full_name }}</li>
    {% endfor %}

Is there anyway to access more than 30? Or what is the common approach for paginating these results? Is it something I do in Liquid or do I need to write some Ruby, etc.?

I am having a hard time finding information about this because "paginating" already has some relevance as Jekyll feature.

Thanks

parkr:
_Should_ fix #16 if I remember correctly.

### DIFF
--- a/lib/jekyll-github-metadata/client.rb
+++ b/lib/jekyll-github-metadata/client.rb
@@ -22,7 +22,7 @@ module Jekyll
           Jekyll.logger.warn "GitHubMetadata:", "No GitHub API authentication could be found." +
             " Some fields may be missing or have incorrect data."
         end
-        Octokit::Client.new(options)
+        Octokit::Client.new({:auto_paginate => false}.merge(options))
       end
 
       def method_missing(meth, *args, &block)


jarrodparkes:
Should auto_paginate be set to true?

...(from <a href="https://github.com/octokit/octokit.rb" target="_blank">docs</a>)

For smallish resource lists, Octokit provides auto pagination. When this is **enabled**, calls for paginated resources will fetch and concatenate the results from every page into a single array:

    Octokit.auto_paginate = true
    issues = Octokit.issues 'rails/rails'
    issues.length

parkr:
@jarrodparkes That must be it!

PR #18, opened 2015-09-04
Enable auto_paginate for Octokit client

### LINKED ISSUE #16: Public Repos (and other collections) only have first 30 results?

parkr:
Fixes #16, replaces #17.

@jarrodparkes Try it now

### DIFF
--- a/lib/jekyll-github-metadata/client.rb
+++ b/lib/jekyll-github-metadata/client.rb
@@ -22,7 +22,7 @@ module Jekyll
           Jekyll.logger.warn "GitHubMetadata:", "No GitHub API authentication could be found." +
             " Some fields may be missing or have incorrect data."
         end
-        Octokit::Client.new(options)
+        Octokit::Client.new({:auto_paginate => true}.merge(options))
       end
 
       def method_missing(meth, *args, &block)


jarrodparkes:
that did it!

jarrodparkes:
might we suggest an option to shut it on/off? or is it possible for me to do that already? (my newbish Ruby) isn't it part of the Gem?

parkr:
This gem is meant to exactly mimic the current github pages metadata feature. It's technically overridable but I want to enforce what GitHub provides :)

jarrodparkes:
<blockquote>It's technically overridable but I want to enforce what GitHub provides :)</blockquote>

With that in mind, how can I override it in my own project (so that you don't have to maintain a separate branch just for me? If this (auto-paginate set to true) can only be set in the Ruby gem, then how can I override it?

Best,
Jarrod

parkr:
> how can I override it in my own project

This, unfortunately, won't be possible. We want it to be one way in all cases! Thanks for spotting the bug though :)

jarrodparkes:
@parkr So what would you suggest as my workaround? Just use another tool/gem?

parkr:
@jarrodparkes If you only want 30, you can limit your loop like this:

```liquid
{% for repo in site.github.repositories limit:30 %}
  <!-- output repo stuff here -->
{% endfor %}
```

PR #19, opened 2015-11-15
Remove build restriction in .travis.yml

Stargator:
There's a restriction in travis that prevents non master branches from triggering a travis build. This causes undo complexity to those that fork the repo and develop a branch to use in the PR.

For instance, I had to create two PRs, one that had the exact change desired and the other that removed the build restriction so I could verify the builds were successful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,4 @@ rvm:
 script: script/cibuild
 notifications:
   email: false
-branches:
-  only:
-    - master
 sudo: false


parkr:
I have this by default because 2 builds for every PR hinders how quickly we can move. Only build for PR is preferable for development. Once the PR is open, every push will trigger a build. We like to encourage "Work In Progress" PR's and this helps with that. :)

Thanks!

Stargator:
Understood.

PR #29, opened 2016-02-05
Add an integration test & test with github-pages gem

parkr:
- [x] integration test to ensure it all works
- [x] conditionally test with github-pages gem
- [x] use webmock so we don't have to worry about making a network call f'realz

/cc @jekyll/gh-pages


jglovier:
img

PR #35, opened 2016-03-01
Add Repository#url_scheme for the pages URL scheme.

### LINKED ISSUE #33: *.github.com domains should have a HTTPS scheme, not HTTP
Perhaps the repo can have an HTTPS flag, which would look to the CNAME, but could be later expanded to look for other conditions.

To note, `github_repo?` is necessary, but not sufficient, as for example, `choosealicense.com` is not HTTPS.

parkr:
Fixes #33.

/cc @benbalter

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -106,7 +106,7 @@ module Jekyll
         end
       end
 
-      def scheme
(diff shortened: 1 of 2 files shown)


benbalter:
Odd that 1/5 tests failed? Otherwise, LGTM (unless you want to put `scheme` on the repository).

Issue #39, opened 2016-03-05
Project sites for users with custom domains return the default domain


benbalter:
I have `benbalter/benbalter.github.com` cname'd to `ben.balter.com`. When used to build `benbalter/benbalter.github.com`, I properly get `http://ben.balter.com` for `site.url`.

If I also have `benbalter/project-site`, with no CNAME, the site is published at `http://ben.balter.com/project-site`, however, `site.url` is generated as `http://benbalter.github.io/project-site`.

For sites without CNAMEs, we'll need to check for `username/username.github.(io|com)`'s cname, if any. It may make sense to update the Page API to return the page's domain (and potentially full URL) in all cases.

PR #40, opened 2016-03-08
Properly determine project page domain by breaking the cache on different args

benbalter:
Right now, if the client makes a call to the same endpoint, but for two different domains (or the same endpoint, but with different arguments), it will return the first, cached result.

That's because the client caches on the method name, irregardless of arguments.

This PR creates a cache hash (rather than using instance variables), and computes a unique key, using SHA1, which I guessed was the cheapest way to generate a unique, consistent hash from the method name + args (knowing that the args may be long).

Thoughts on this approach?

Fixes https://github.com/jekyll/github-metadata/issues/39

### DIFF
--- a/lib/jekyll-github-metadata.rb
+++ b/lib/jekyll-github-metadata.rb
@@ -1,5 +1,4 @@
 require 'octokit'
-require 'digest'
 
 module Jekyll
   unless const_defined? :Errors
--- a/lib/jekyll-github-metadata.rb
+++ b/lib/jekyll-github-metadata.rb
@@ -1,4 +1,5 @@
 require 'octokit'
+require 'digest'
 
 module Jekyll
   unless const_defined? :Errors
--- a/lib/jekyll-github-metadata/client.rb
+++ b/lib/jekyll-github-metadata/client.rb
@@ -45,10 +45,9 @@ module Jekyll
       def method_missing(method_name, *args, &block)
         method = method_name.to_s
         if accepts_client_method?(method_name)
-          instance_var_name = method.sub('?', '_')
+          key = cache_key(method_name, args)
           Jekyll.logger.debug "GitHub Metadata:", "Calling @client.#{method}(#{args.map(&:inspect).join(", ")})"
-          instance_variable_get(:"@#{instance_var_name}") ||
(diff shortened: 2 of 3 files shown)


lib/jekyll-github-metadata/client.rb
```diff
@@ -45,10 +45,9 @@ def respond_to?(method_name, include_private = false)
       def method_missing(method_name, *args, &block)
         method = method_name.to_s
         if accepts_client_method?(method_name)
-          instance_var_name = method.sub('?', '_')
+          key = cache_key(method_name, args)
           Jekyll.logger.debug "GitHub Metadata:", "Calling @client.#{method}(#{args.map(&:inspect).join(", ")})"
-          instance_variable_get(:"@#{instance_var_name}") ||
-            instance_variable_set(:"@#{instance_var_name}", save_from_errors { @client.public_send(method_name, *args, &block) })
+          cache[key] || cache[key] = save_from_errors { @client.public_send(method_name, *args, &block) }
```
parkr: Let's do `cache[key] ||= save_from_errors ...` here.

lib/jekyll-github-metadata.rb
```diff
@@ -1,4 +1,5 @@
 require 'octokit'
+require 'digest'
```
parkr: I'm starting to think each file should have its own requires. They're aggressively cached anyway so if there's any duplication, we should be fine. This would mean we could require a file by itself and it'd pull in all the files it needs.

parkr:
CI failures are because we don't have stubbed requests for the relevant new calls/.

benbalter:
@parkr All makes sense. I'm not going to be able to tackle this for a few hours, if you have a chance to first, and want to get it out, feel free to hijack my branch, otherwise, I should have a chance to this afternoon.

lib/jekyll-github-metadata/client.rb
benbalter: Dumb move on my part. Nice call.

benbalter:
@parkr ready for another round of :eyes:

parkr:
:heart: 

@jekyllbot: merge

PR #42, opened 2016-03-16
Mark the generator as safe so in safe mode it'll work

parkr:
When debugging output would be helpful...

[If the plugin isn't safe, it won't be included in the set](https://github.com/jekyll/jekyll/blob/v3.0.3/lib/jekyll/site.rb#L292-L305).

/cc @jekyll/gh-pages


parkr:
@jekyllbot: merge

Issue #43, opened 2016-03-17
Default PAGES_ENV should be development, not dotcom


benbalter:
So that when run locally, with no additional configuration, it works as expected, and in production, a more sophisticated environment, where it's easier to pass configuration variables, it can be overridden.

parkr:
> it works as expected

What is broken?

benbalter:
Right now, by default, when I run this plugin locally with no additional configuration flags it says it's "dotcom". It should say that it's "development"

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If you can still reproduce this error on the <pre>3.1-stable</pre> or <pre>master</pre> branch,
please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider building it first as a plugin. Jekyll 3 introduced
[hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout
the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be
built as a plugin, then please provide more information about why in order to keep this issue open.

Thank you for all your contributions.

benbalter:
I'd still like to see/discuss this along with #49

Issue #48, opened 2016-03-30
Make it easier to use GitHub metadata with GitHub Enterprise


benbalter:
Right now, if I want to use this plugin with GitHub Enterprise, it looks like I need to set the following environmental variables:

* `JEKYLL_GITHUB_TOKEN` - unavoidable
* `SSL` - this should be detected from another variable's scheme
* `OCTOKIT_WEB_ENDPOINT` -
* `OCTOKIT_API_ENDPOINT` and `PAGES_API_URL` -
* `PAGES_PAGES_HOSTNAME` and `PAGES_GITHUB_HOSTNAME` - oddly named, should be determined from Enterprise hostname
* `PAGES_ENV` - should be development, should not require "enteprise" as it does now

Not documented, but I may also need to set:

* `SUBDOMAIN_ISOLATION`

Instead, I should set:

* `JEKYLL_GITHUB_TOKEN`
* `PAGES_ENDPOINT` (or something similar)
* (and if necessary) `SUBDOMAIN_ISOLATION`

From there, the plugin should pull `SSL` from the scheme, set Octokit's endpoint, know the API endpoint `/api/v3/`, know the Pages hostname, and should default the env to development.

In any event, right now, it's really painful to use this Gem with Enterprise, so the documentation could use a bit of :heart:.

parkr:
> `PAGES_ENDPOINT`

This is basically your `ENTERPRISE_URL`, aka `https://github.acme.com`?

benbalter:
Yep. That was my proposal.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
three months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If you can still reproduce this error on the <pre>3.1-stable</pre> or <pre>master</pre> branch,
please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider building it first as a plugin. Jekyll 3 introduced
[hooks](http://jekyllrb.com/docs/plugins/#hooks) which provide convenient access points throughout
the Jekyll build pipeline whereby most needs can be fulfilled. If this is something that cannot be
built as a plugin, then please provide more information about why in order to keep this issue open.

Thank you for all your contributions.

PR #49, opened 2016-03-30
Default to development in dev

benbalter:
Fixes https://github.com/jekyll/github-metadata/issues/43.

Also fixes the domain to use the Pages hostname, not the GitHub hostname, when custom domains are disabled in order to get tests to pass.


parkr:
@benbalter Isn't that supposed to be `JEKYLL_ENV`? `PAGES_ENV` is used to determine the target to request from, I think.

benbalter:
> PAGES_ENV is used to determine the target to request from, I think.

That doesn't make much sense to me. The idea behind `PAGES_ENV` was (A) so that we could know if we were in enterprise or dotcom internally, and (B) so that the user could know if they were in dotcom (or enterprise) or locally. I believe it predates `JEKYLL_ENV`. Additionally, is `JEKYLL_ENV` exposed to liquid where the user can access it? 

If we thought of using `PAGES_ENV` as a flag/toggle, it'd really only be useful, if I could set `PAGES_ENV` to the enterprise instance, and everything else just worked. Otherwise, I can't see when a user would want to actively toggle back and forth. Even if pushing to both remotes, they'd want to render things differently locally (e.g., changing the URL, not including discuss, etc.).

In my mind:

* `PAGES_ENV`: should be `dotcom`, `enterprise`, or `development` (default)
* `JEKYLL_ENV: should be `production` or `development`

The two are similar, but `PAGES_ENV` is exposed to liquid, and can be used to differentiate between GitHub environments.

parkr:
@benbalter `jekyll.environment` exposes `JEKYLL_ENV`. It too defaults to `development`. The only difference I see is that `PAGES_ENV` differentiates between `dotcom` and `enterprise`, which are both `production` for `JEKYLL_ENV`.

What should `PAGES_ENV=development` do about changing the octokit API calls? `https://github.dev`?

benbalter:
> What should PAGES_ENV=development do about changing the octokit API calls? https://github.dev?

I think nothing? As someone previewing my Pages-hosted site locally, I'd want the metadata (save the URL) to be the live metadata, no?

parkr:
> I think nothing? As someone previewing my Pages-hosted site locally, I'd want the metadata (save the URL) to be the live metadata, no?

What if it's an Enterprise repo? Wouldn't that affect this? Would we have `development` and `development-enterprise`?

benbalter:
> What if it's an Enterprise repo?

That's what I was asking about in https://github.com/jekyll/github-metadata/issues/48. (and suggested ENDPOINT)

lib/jekyll-github-metadata/repository.rb
```diff
@@ -172,7 +172,7 @@ def cname
       def domain
         @domain ||=
           if !Pages.custom_domains_enabled?
-            Pages.github_hostname
+            Pages.pages_hostname
```
parkr: An associated test may need to be updated for this.

benbalter:
@parkr got tests passing, if you're able to :eyes: once more. :smile: Would be good to finally get this out.

parkr:
LGTM.

benbalter:
LGTM.

benbalter:
@jekyllbot: merge +major

PR #53, opened 2016-03-31
Load ENV via dotenv if available

Crunch09:
Hi,
i didn't know if tests are needed for this because i think they would only test the behavior already tested in the dotenv repository. Also i'm not sure if i should add something to the README about this.

### DIFF
--- a/lib/jekyll-github-metadata.rb
+++ b/lib/jekyll-github-metadata.rb
@@ -1,4 +1,11 @@
 require 'octokit'
+if defined?(Jekyll) && Jekyll.respond_to?(:env) && Jekyll.env == 'development'
+  begin
+    require 'dotenv'
+    Dotenv.load
+  rescue LoadError
+  end
+end
 
 module Jekyll
   unless const_defined? :Errors


lib/jekyll-github-metadata.rb
```diff
@@ -1,4 +1,9 @@
 require 'octokit'
+begin
+  require 'dotenv'
```
benbalter: What do you think about wrapping these two lines (also) in a `if Jekyll.env == 'development'`? 

I believe we'll also need to `require 'jekyll'` then, which I don't *think* will cause any problems.

Crunch09: Sure, sounds good! But i think we should put `require 'jekyll'` also in the  `begin` block  then because it was removed in https://github.com/jekyll/github-metadata/commit/80370fc0c04198bc6e5e3cdce4966e9a17dbbb73 and it's only a development dependency in the gemspec

benbalter: The other way *may* be:

```ruby
if defined? Jekyll && Jekyll.env == "development"
```

Crunch09:
@benbalter Hi, i updated the PR as we discussed.

benbalter:
@parkr what do you think?

parkr:
@benbalter I THINK GOOOOOOD!

benbalter:
Thanks @Crunch09! 🌷 🐢 🎉

PR #55, opened 2016-04-06
Refactor some of the git things & better docs

parkr:
@benbalter @jekyll/gh-pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,18 @@ Usage of this gem is pretty straight-forward. Add it to your bundle like this:
 gem 'jekyll-github-metadata'
 ```
 
-Then go ahead and run `bundle install`. Once you've done that, add your repo & the gem to your `_config.yml`:
+Then go ahead and run `bundle install`. Once you've done that, add your repo name & the gem to your `_config.yml`:
 
 ```yaml
-# or use PAGES_REPO_NWO in the env
+# set PAGES_REPO_NWO=me/super-cool-project in the env
 repository: me/super-cool-project
 gems: ['jekyll-github-metadata']
 ```
+
+"NWO" stands for "name with owner." It is GitHub lingo for the username of
+the owner of the repository plus a forward slash plus the name of the
+repository, e.g. 'parkr/blog', where 'parkr' is the owner and 'blog' is the
+repository name.
 
 Then run `jekyll` like you normally would and your `site.github.*` fields should fill in like normal.
 
--- a/README.md
+++ b/README.md
@@ -13,12 +13,32 @@ Usage of this gem is pretty straight-forward. Add it to your bundle like this:
 gem 'jekyll-github-metadata'
 ```
 
-Then go ahead and run `bundle install`. Once you've done that, add your repo name & the gem to your `_config.yml`:
+Now add it to your `_config.yml`:
 
 ```yaml
-# set PAGES_REPO_NWO=me/super-cool-project in the env
-repository: me/super-cool-project
 gems: ['jekyll-github-metadata']
+```
+
+Then go ahead and run `bundle install`. Once you've done that jekyll-github-metadata will run when you run Jekyll.
+
+In order for jekyll-github-metadata to know what metadata to fetch it must
+be able to determine the repository to ask GitHub about.
+
+The easiest way to accomplish this is by setting an "origin" remote with a
+github.com URL. If you run `git remote -v` in your repository, you should
+see your repo's URL.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,15 +13,42 @@ Usage of this gem is pretty straight-forward. Add it to your bundle like this:
...
 gems: ['jekyll-github-metadata']
 ```
 
-Then run `jekyll` like you normally would and your `site.github.*` fields should fill in like normal.
+Then go ahead and run `bundle install`. Once you've done that jekyll-github-metadata will run when you run Jekyll.
+
+In order for jekyll-github-metadata to know what metadata to fetch it must
+be able to determine the repository to ask GitHub about.
+
+The easiest way to accomplish this is by setting an "origin" remote with a
```
benbalter: I'd note that if you've run `git clone` from GitHub, this is almost 100% set, no further action needed.

README.md
```diff
@@ -13,15 +13,42 @@ Usage of this gem is pretty straight-forward. Add it to your bundle like this:
...
+github.com URL. If you run `git remote -v` in your repository, you should
+see your repo's URL.
+
+Alternatively, you can set the repo NWO as either the `PAGES_REPO_NWO`
+environment variable:
```
benbalter: > Alternatively

When would you need this?

README.md
```diff
@@ -13,15 +13,42 @@ Usage of this gem is pretty straight-forward. Add it to your bundle like this:
...
+
+```shell
+$ PAGES_REPO_NWO=username/repo-name bundle exec jekyll build
+```
+
+... or add your repo name to your `_config.yml`:
```
benbalter: Why three options? Why is the ENV var before the config (which users presumably now)?

lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -34,19 +34,39 @@ def git_remote_url
...
       end
 
+      def nwo_from_env
+        ENV['PAGES_REPO_NWO']
+      end
+
+      def nwo_from_config(site)
+        nwo = site.config['repository']
+        nwo if nwo && nwo.is_a?(String) && nwo.include?('/')
+      end
```
benbalter: The three methods above.... really like this clean implementation.

lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -34,19 +34,39 @@ def git_remote_url
...
-          git_nwo || \
+        nwo_from_env || \
+          nwo_from_config || \
+          nwo_from_git_origin_remote || \
           proc {
-            raise GitHubMetadata::NoRepositoryError, "No repo name found. "
-              "Specify using PAGES_REPO_NWO or 'repository' in your configuration."
+            raise GitHubMetadata::NoRepositoryError, "No repo name found. " \
+              "Specify using PAGES_REPO_NWO environment variables, " \
+              "'repository' in your configuration, or set up an 'origin' git remote."
```
benbalter: To note, origin must be set to `github.com`

lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -34,19 +34,39 @@ def git_remote_url
         end.uniq.first || ""
       end
 
-      def git_nwo
+      def nwo_from_git_origin_remote
         return unless Jekyll.env == "development"
         matches = git_remote_url.match %r{github.com(:|/)([\w-]+)/([\w-]+)}
```
benbalter: To note, this will only work with .com, not Enterprise. Maybe we should ignore the host?

README.md
parkr: I think the config is a bad option as it sits now. `repository` could be a number of things other than "the NWO corresponding to my github repo for my project". I think it'll cause the most issues, so I put it last.

README.md
parkr: When they don't have the git remote available.

lib/jekyll-github-metadata/ghp_metadata_generator.rb
parkr: @benbalter What if we use [`Pages. github_hostname`](https://github.com/jekyll/github-metadata/blob/5e38433ca027efbb3491c6e04e8a844cdc5b56e6/lib/jekyll-github-metadata/pages.rb#L59-L61)

parkr:
@jekyllbot: merge

PR #56, opened 2016-04-22
Make the Client::API_CALLS a Set

parkr:
Avoid duplicates. Also, unfreeze it so we can add things if necessary.


parkr:
@jekyllbot: merge

PR #58, opened 2016-05-13
Add site.github.private_repositories field

### LINKED ISSUE #23: possibility to use private repo's
This would be interesting for repo's that have a `gh-pages` branch, but are private, or for personal use.

This would use the `visibility` key present in the api.

https://developer.github.com/v3/repos/

Crunch09:
This creates a GET request to /user/repos with type set to `private`.

For the webmock in `api_get_accessible_private_repos.json` i used the same json as `api_get_owner_repos.json` only with the `private` attributes set to `true`.

fixes #23


benbalter:
There's a security concern here, although I'm not sure how likely or how large. 

GitHub Pages sites are built with the pusher's OAuth token. Adding this endpoint could create an existence disclosure vulnerability, in which the name of and metadata regarding private repos are published inadvertently. It'd require the repo collaborator to trigger a build (e.g., on merge), but there might be cases, e.g., branch builds that that's not true. Not saying no, just saying we need to think through and document the implications.

Put another way, what's the use case for wanting to disclose the existence of private repos programmatically?

Crunch09:
Hi @benbalter ,

thanks for your quick feedback! You're right, there could be a security issue if an attacker is able to access the result of `private_repositories` or at least the API call. The use case was #23 but maybe i didn't understand the feature request correctly.

PR #59, opened 2016-05-13
remove unneeded param `type`

Crunch09:
Hi,
`GET /users/:username/repos` only returns public repos by default and `public` isn't even one of the available options for the `type` param.
See https://developer.github.com/v3/repos/#list-user-repositories for more
information.

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -69,7 +69,7 @@ module Jekyll
       end
 
       def owner_public_repositories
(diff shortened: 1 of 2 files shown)


parkr:
@benbalter Oh, this needed `type => public` before... did this recently change?

parkr:
I think we meant to do `/user/repos`: https://developer.github.com/v3/repos/#list-your-repositories

Crunch09:
@parkr Yes, i think it's good that we don't use `/user/repos` because as of now the user doesn't need to be authenticated if he / she only wants to access public fields and this would change if we used that call.

benbalter:
Should this be `visibility`?

PR #60, opened 2016-05-25
Use html_url from Pages endpoint (behind preview env flag)

parkr:
Adds a `RepositoryCompat` class which has a fallback `#pages_url` that `Repository#pages_url` uses in case it doesn't get the `/pages` info.

- [x] Handle preview API
- [x] Make it opt-in

/cc @jekyll/gh-pages for review.


benbalter:
@parkr having a little trouble following the diff, due to noise. To confirm, it should use the API value for all fields (url, scheme, domain, etc.), and fall back to its own calculation, if the API isn't available, right?

lib/jekyll-github-metadata/repository.rb
```diff
@@ -8,14 +8,18 @@ def initialize(name_with_owner)
...
+        @repo_compat ||= Jekyll::GitHubMetadata::RepositoryCompat.new(self)
       end
 
       def repo_info
         @repo_info ||= (Value.new(proc { |c| c.repository(nwo) }).render || Hash.new)
       end
 
+      def repo_pages_info
+        @repo_pages_info ||= (Value.new(proc { |c| c.pages(nwo) }).render || Hash.new)
+      end
```
parkr: @benbalter this is key.

lib/jekyll-github-metadata/repository.rb
```diff
@@ -130,56 +130,26 @@ def default_user_domain
...
-          else
-            URI.join("#{Pages.github_url}/pages/", path).to_s
-          end
-        elsif cname || primary?
-          "#{url_scheme}://#{domain}"
-        else
-          URI.join("#{url_scheme}://#{domain}", name).to_s
-        end
+      def html_url
+        @html_url ||= (repo_pages_info["html_url"] || repo_compat.pages_url)
```
parkr: @benbalter as well as this. this is where it grabs the `html_url` from the pages endpoint if it exists, otherwise uses the old `pages_url` like before.

parkr:
> To confirm, it should use the API value for all fields (url, scheme, domain, etc.), and fall back to its own calculation, if the API isn't available, right?

It uses the API value for all fields (taking the `html_url`, putting it into `URI()` and extracting from there) unless the API value doesn't exist. If it doesn't exist, then it proxies to the RepoCompat layer, which is the old behaviour.

benbalter:
Thanks for the clarification. Makes sense to me.

parkr:
@benbalter Confirmed that the current code is opt-in via `PAGES_PREVIEW_HTML_URL` and works fine on our canary page.

parkr:
@jekyllbot: merge +minor

PR #64, opened 2016-06-22
Remove trailing slash from html_url if present

benbalter:
For consistency, the `html_url` field of the pages preview API will begin returning URLs with trailing slashes in all cases. Because many sites hardcode things like `{{ page.url | prepend: site.github.url }}`, we need to maintain backwards compatibility, and have `site.github.url` continue to *not* have a trailing slash.

This PR simply strips the trailing slash from the calculated URL or API response, if it's present, so that we never have a trailing slash. Ever.

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -144,7 +144,7 @@ module Jekyll
       end
 
       def html_url
-        @html_url ||= (repo_pages_info["html_url"].chomp("/") || repo_compat.pages_url)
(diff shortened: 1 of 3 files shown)


benbalter:
@parkr how does this look to you? Looks like tests are failing because 2.0.1 was released, but not in the latest version of the Pages gem.

I'm going to suggest we merge this here, cut 2.0.2, and bump the pages gem once with both, since 2.0.1 contains no in-production changes.

parkr:
LGTM. I need to figure out a better way to test against the `github-pages` gem. Maybe just forcing the `version.rb` file to contain whatever `github-pages` depends on...

PR #65, opened 2016-07-12
Expose mention URL base

benbalter:
Part of the fix for https://github.com/jekyll/jekyll-mentions/issues/34, this pull request exposes `site.github.mention_url_base`, which will be the hostname + scheme, to allow Jekyll mentions to default to the proper hostname in Enterprise.

### DIFF
--- a/lib/jekyll-github-metadata/metadata_drop.rb
+++ b/lib/jekyll-github-metadata/metadata_drop.rb
@@ -23,6 +23,7 @@ module Jekyll
       def_delegator :"Jekyll::GitHubMetadata::Pages", :pages_hostname, :pages_hostname
       def_delegator :"Jekyll::GitHubMetadata::Pages", :api_url, :api_url
       def_delegator :"Jekyll::GitHubMetadata::Pages", :help_url, :help_url
+      def_delegator :"Jekyll::GitHubMetadata::Pages", :mention_url_base
 
       def versions
         @versions ||= begin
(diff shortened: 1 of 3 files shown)


lib/jekyll-github-metadata/metadata_drop.rb
```diff
@@ -23,6 +23,7 @@ def keys
       def_delegator :"Jekyll::GitHubMetadata::Pages", :pages_hostname, :pages_hostname
       def_delegator :"Jekyll::GitHubMetadata::Pages", :api_url, :api_url
       def_delegator :"Jekyll::GitHubMetadata::Pages", :help_url, :help_url
+      def_delegator :"Jekyll::GitHubMetadata::Pages", :mention_url_base
```
jdennes: Just a naming thing, but `mention_base_url` might be a bit more consistent with the other URLs defined here.

jdennes: Just a naming thing, but `mention_base_url` might be a bit more consistent with the other URLs defined above.

parkr: True ^ Also, `def_delegator` needs three arguments-- you only have 2 there :)

benbalter: It only requires two [if the alias and the delegated method are the same](http://ruby-doc.org/stdlib-2.0.0/libdoc/forwardable/rdoc/Forwardable.html#method-i-def_delegator)

parkr: Hm, ok. TIL! Would you mind just adding it for consistency? All the ones above could be refactored in a later PR to only use the exact number of args they require.

jdennes:
Think this can be closed now.

PR #69, opened 2016-08-11
Add rubocop

benbalter:
Based off of Jekyll's Rubocop config.

### DIFF
--- a/lib/jekyll-github-metadata/client.rb
+++ b/lib/jekyll-github-metadata/client.rb
@@ -57,12 +57,8 @@ module Jekyll
         end
       end
 
-      def save_from_errors(default = false, &block)
-        if block.arity == 1
-          yield(@client)
-        else
-          yield
-        end
+      def save_from_errors(default = false)
+        yield @client
       rescue Faraday::Error::ConnectionFailed, Octokit::TooManyRequests => e
         Jekyll::GitHubMetadata.log :warn, e.message
         default
--- a/lib/jekyll-github-metadata/client.rb
+++ b/lib/jekyll-github-metadata/client.rb
@@ -1,4 +1,4 @@
-require 'digest'
+require "digest"
 
 module Jekyll
   module GitHubMetadata
@@ -6,7 +6,7 @@ module Jekyll
       InvalidMethodError = Class.new(NoMethodError)
 
       # Whitelisted API calls.
-      API_CALLS = Set.new(%w{
+      API_CALLS = Set.new(%w(
         repository
         organization
         repository?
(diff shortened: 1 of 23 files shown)


lib/jekyll-github-metadata/client.rb
```diff
@@ -59,9 +59,9 @@ def method_missing(method_name, *args, &block)
 
       def save_from_errors(default = false, &block)
         if block.arity == 1
-          block.call(@client)
+          yield(@client)
         else
-          block.call
+          yield
```
parkr: wont this be a problem? What happens to `block` here? Is yield supposed to know to call block? :confused:

script/test
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
 echo Testing with $(bundle exec jekyll --version)
 bundle exec rspec --color --require spec_helper $@
+bundle exec rubocop -S -D
+bundle exec gem build jekyll-github-metadata.gemspec
```
parkr: These two actions seem unrelated to rspec. Perhaps they should go in `script/cibuild` instead?

spec/spec_helper.rb
```diff
@@ -149,3 +149,10 @@ def env_args_to_hash(*args)
     Jekyll::GitHubMetadata.logger = Logger.new(StringIO.new) unless ENV["DEBUG"]
   end
 end
+
+def versions
+  require "github-pages"
+  GitHubPages.versions
+rescue LoadError
+  {}
+end
```
parkr: Why do this?

parkr:
Thank you! :heart: a couple comments above if you wouldn't mind clarifying.

script/test
benbalter: Good call. Will do.

spec/spec_helper.rb
benbalter: Rubocop complained that it didn't like the original chained call, as it was sandwiched in a lamda inside a hash which ended with `end.call`. This was my best attempt to preserve the original behavior while not angering Rubocop.

parkr: Ah ok. We have removed the test with `github-pages` so maybe remove this for now with a comment?

lib/jekyll-github-metadata/client.rb
benbalter: That was Rubocop's doing, and tests pass?

parkr: Yeah, it's just confusing 😜 I *think* `yield` doesn't care about the arity of the block so maybe just simplify this to 

```ruby
def save_from_errors(default = false)
  yield @client
rescue # ...
# ...
end
```

PR #70, opened 2016-08-11
Dont try to build with the latest released pages gem to avoid conflicts

benbalter:
Tests on master are currently broken, and given our current set up, will always be broken when this repo has a different version number than the version used in the last release of GitHub Pages.

This PR updates the travis config to not try to build against the latest Pages gem version (which is built as part of CI upstream now).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: ruby
 rvm:
   - 2.1
   - 2.2
-  - 2.3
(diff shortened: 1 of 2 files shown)


benbalter:
LGTM.

Going to merge since this simply makes tests green on `master`, and can go back and re-add the `github-pages` test, if there's value and it's possible to implement in a way that doesn't break.

PR #75, opened 2016-10-04
If a user provides bad credentials, throw an error.

### LINKED ISSUE #67: Fail loudly with an invalid token

parkr:
This fixes #67. If you give me `ACCESS_TOKEN=1234`, you'll receive a loud error when you run `jekyll serve` that it's invalid. Providing either no token or a valid token will solve the error.

/cc @jekyll/gh-pages for review.


lib/jekyll-github-metadata/client.rb
```diff
@@ -71,13 +72,25 @@ def inspect
         "#<#{self.class.name} @client=#{client_inspect}>"
       end
 
+      def authenticated?
+        !@client.access_token.nil? && !@client.access_token.empty?
```
benbalter: Is this better as `to_s.empty?`?

lib/jekyll-github-metadata/client.rb
```diff
@@ -71,13 +72,25 @@ def inspect
...
+      end
+
+      # Raise an error if credentials are provided and they cause a 401.
+      def check_credentials!
+        return unless authenticated?
+        @client.user
```
benbalter: Why make an extra call to check credentials, when we can just rescue if the call we need fails due to bad auth?

lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -8,6 +8,9 @@ class GHPMetadataGenerator < Jekyll::Generator
       def generate(site)
         Jekyll::GitHubMetadata.log :debug, "Initializing..."
 
+        # Fail loudly if credentials are provided but aren't valid.
+        Jekyll::GitHubMetadata.client.check_credentials!
```
benbalter: Won't this be a change in behavior? Right now, I can have an invalid credential, but if I don't use any `site.github` metadata, the drop isn't called, and thus it's okay, right? I don't know that we need to go out of our way to fail a user's build if they're not relying on the plugin for their site to build.

parkr:
@benbalter Ready for another round of review!

lib/jekyll-github-metadata/client.rb
```diff
@@ -15,6 +16,7 @@ class Client
         releases
         list_repos
         organization_public_members
+        user
```
benbalter: Do we need this still?

benbalter:
@parkr my bad. One question, which I think is left over from the previous iteration. 

Can this go in a point release, or does it need its own minor release?

parkr:
@benbalter OK. Ready for another review.

parkr:
@jekyllbot: merge +minor

PR #76, opened 2016-10-05
Spike out setting site.url and site.baseurl

benbalter:
This allows `relative_url` and `absolute_url` to "just work" in most cases, by silently setting `site.url` and `site.baseurl` if either are not already set by the user.

This is a quick spike, and could use:

- [x] Code review
- [x] Passing tests

### DIFF
--- a/lib/jekyll-github-metadata/ghp_metadata_generator.rb
+++ b/lib/jekyll-github-metadata/ghp_metadata_generator.rb
@@ -5,18 +5,31 @@ module Jekyll
     class GHPMetadataGenerator < Jekyll::Generator
       safe true
 
+      attr_reader :site
+
       def generate(site)
         Jekyll::GitHubMetadata.log :debug, "Initializing..."
+        @site = site
+        site.config["github"]  = github_namespace
+        site.config["url"]     = drop.url if site.config["url"].nil?
+        site.config["baseurl"] = drop.baseurl if site.config["baseurl"].nil?
+      end
 
-        site.config["github"] =
-          case site.config["github"]
-          when nil
-            MetadataDrop.new(site)
-          when Hash, Liquid::Drop
-            Jekyll::Utils.deep_merge_hashes(MetadataDrop.new(site), site.config["github"])
-          else
-            site.config["github"]
-          end
+      private
+
+      def github_namespace
+        case site.config["github"]
+        when nil
+          drop
+        when Hash, Liquid::Drop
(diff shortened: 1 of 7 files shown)


benbalter:
@parkr I believe you have a similar implementation... if you want to take this over, or close in favor of yours, please feel free to do so if you think you can get it across the 🏁 .

lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -5,18 +5,31 @@ module GitHubMetadata
     class GHPMetadataGenerator < Jekyll::Generator
       safe true
 
+      attr_reader :site
+
       def generate(site)
         Jekyll::GitHubMetadata.log :debug, "Initializing..."
+        @site = site
+        site.config["github"]  = github_namespace
+        site.config["url"]     = drop.url if site.config["url"].nil?
```
parkr: This could be `site.config["url"] ||= drop.url` and same for `baseurl`. Would save you an explicit `if` 😄

spec/ghp_metadata_generator_spec.rb
```diff
@@ -4,8 +4,61 @@
...
     expect(described_class.safe).to be(true)
   end
+
+  context "generating" do
+    before { stub_api("/repos/jekyll/github-metadata" , "repo") }
+    before { stub_api("/repos/jekyll/github-metadata/pages" , "repo_pages") }
+    before { stub_api("/repos/jekyll/jekyll.github.io" , "repo") }
+    before { stub_api("/repos/jekyll/jekyll.github.io/pages" , "repo_pages") }
+    before { stub_api("/repos/jekyll/jekyll.github.com" , "repo") }
+    before { stub_api("/repos/jekyll/jekyll.github.com/pages" , "repo_pages") }
```
parkr: I'd put all of these into one `before(:each)` block. Should be a bit faster. Same with the env vars below.

benbalter:
@parkr made the `||=` change... if you have time to help get the tests over the finish line, I think you're more knowledgeable about the tests here (which is where I started hitting my limits).

parkr:
@benbalter Got them working on my machine. Does this LGTY?

parkr:
@benbalter Because you made the PR you can't approve or request changes, but feel free to comment and I'll check back for your final 👍 .

benbalter:
@parkr Nice work. Thanks for the help with the tests. Really like your implementation (which is where I was getting stuck). Think this is going to be a big improvement. ✨

PR #78, opened 2016-10-11
Add MetadataDrop#to_s which outputs pretty JSON representation

parkr:
This fixes an issue where `{{ site.github }}` doesn't expand out to all the values in the Drop.

This is controlled by `Drop#to_s` which, by default, is `self.class.name`.

/cc @jekyll/gh-pages

### DIFF
--- a/spec/metadata_drop_spec.rb
+++ b/spec/metadata_drop_spec.rb
@@ -2,9 +2,25 @@ require "spec_helper"
 
 RSpec.describe(Jekyll::GitHubMetadata::MetadataDrop) do
   let(:overrides) { { "repository" => "jekyll/another-repo" } }
-  let(:config) { Jekyll::Configuration::DEFAULTS.merge(overrides) }
+  let(:config) { Jekyll::Configuration.from(overrides) }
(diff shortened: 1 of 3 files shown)


spec/metadata_drop_spec.rb
```diff
@@ -2,10 +2,26 @@
...
+      site.config.delete("repository")
+      site.config["github"] = subject
+    end
+
+    it "renders as-is in Liquid" do
+      payload = site.site_payload
+      expect(payload["site"]["github"]).to be_instance_of(described_class)
+      template = Liquid::Template.parse("{{ site.github }}")
+      result = template.render!(payload, :registers => { :site => site })
+      expect(result).to eql(subject.to_s)
```
benbalter: Aren't you just testing that Liquid calls `to_s` on this? Wouldn't this be true if the `to_s` method wasn't defined? Perhaps match a regex on the content?

parkr: Good point. Updated to match the `JSON.pretty_generate` which is already tested.

PR #79, opened 2016-10-24
Lock Octokit to v4.3.0

parkr:
v4.4.0 introduced a change where it requires preview headers for `GET /repos/o/r/pages`, which aren't required. This will break our requests to get Pages metadata, such as `cname`, `html_url`, etc.

PR which breaks it: https://github.com/octokit/octokit.rb/pull/815
PR which fixes it: https://github.com/octokit/octokit.rb/pull/822

/cc @jekyll/gh-pages

PR #81, opened 2016-10-25
Revert "Lock Octokit to v4.3.0", but disallow v4.4.0

parkr:
Reverts jekyll/github-metadata#79, but allows octokit v4.4.1 which has the fix (https://github.com/octokit/octokit.rb/pull/822).

https://github.com/octokit/octokit.rb/releases/tag/v4.4.1

/cc @jekyll/gh-pages


parkr:
@jekyllbot: merge +bug

PR #84, opened 2016-12-19
Make the Octokit client more configurable.

parkr:
At the moment, it only points to https://api.github.com, but it should be configurable for GitHub Enterprise to point to other API and Web endpoints.

/cc @jekyll/gh-pages @brandonblack


parkr:
@benbalter By setting the relevant environment variables.

benbalter:
Ahh, so what's descibed in https://github.com/jekyll/github-metadata#overrides isn't accurate today?

parkr:
@benbalter if you set PAGES_API_URL, octokit wont use it without this patch. You *also* have to set the OCTOKIT_API_ENDPONT environment variable.

benbalter:
@jekyllbot: merge

PR #85, opened 2017-01-02
Respect source passed from the API

benbalter:
The Pages API now exposes source branch and path metadata. We should use that to build the proper zip and tar URLs so we respect the source selector.


benbalter:
If you don't mind?

parkr:
@jekyllbot: merge +minor

PR #88, opened 2017-01-24
add `latest_release` and `latest_release_url`

### LINKED ISSUE #80: Add site.github.releases.latest
Sometimes a release is tagged for in the past for historical purposes.  Grabbing all of the releases and hoping that the first one is the latest doesn't work.  Instead, it could query `/repos/:owner/:repo/releases/latest`<sup>1</sup> which will give the release marked latest.

[1] https://developer.github.com/v3/repos/releases/#get-the-latest-release

Crunch09:
Hi,
this uses [Octokit::Client::Releases#latest_release](https://github.com/octokit/octokit.rb/blob/adab5f6cc2e4c896df64a43f4ce0711398631393/lib/octokit/client/releases.rb#L145) and fixes #80 .


lib/jekyll-github-metadata/repository.rb
```diff
@@ -64,6 +64,10 @@ def releases_url
         "#{repository_url}/releases"
       end
 
+      def latest_release_url
+        "#{releases_url}/latest"
```
benbalter: Could this be `latest_release.url`?

Crunch09: Do you mean remove `latest_release_url` because as i just saw `url` is already included in the payload from github so the user could write `{{ site.github.latest_release.url }}` ? Or do you mean calling (if possible) `latest_release.url` instead of `"#{releases_url}/latest"` on this line?

benbalter: > Or do you mean calling (if possible) latest_release.url instead of "#{releases_url}/latest" on this line?

This. Meaning can we call the `url` method of the `latest_release` object, rather than constructing the URL ourselves?

Crunch09: Thank you for the clarification! Just updated to use the `latest_release` object (hash).

benbalter:
@jekyllbot: merge

benbalter:
Thanks @Crunch09! Sorry for the delay in merging!

Crunch09:
No worries! Thank you 👍

PR #89, opened 2017-01-24
Add .configuration and .page_build? methods to Pages

parkr:
`.configuration` gives us a way to debug our metadata fetching by showing what the Pages configuration is.

`.page_build?` gives us a different way to enable automatically setting the `site.url` and `site.baseurl` values when running in development mode.

/cc @jldec


parkr:
@jekyllbot: merge +minor

PR #92, opened 2017-02-06
Add instructions for using DotEnv

njt1982:
Follow up to #53 - lets people know you can use a `.env` file... (Seems to work for me?)


parkr:
@jekyllbot: merge +docs

ashmaroli:
@jekyllbot: merge +docs

PR #94, opened 2017-02-22
Add commits to site.github

jerrylopez:
I'd like to propose that commits be added to site.github. The main reason for this would be for display your commit history on your generated site.

Currently I am building an internal documentation site and thought it would be great to have a recent activity feed or history feed.

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -98,6 +98,10 @@ module Jekyll
         memoize_value :@releases, Value.new(proc { |c| c.releases(nwo) })
       end
 
+      def commits
+        memoize_value :@commits, Value.new(proc { |c| c.commits(nwo) })
+      end
+
       def git_ref
         if repo_pages_info["source"]
           repo_pages_info["source"]["branch"]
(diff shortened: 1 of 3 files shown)


lib/jekyll-github-metadata/repository.rb
```diff
@@ -98,6 +98,10 @@ def releases
         memoize_value :@releases, Value.new(proc { |c| c.releases(nwo) })
       end
 
+      def commits
+        memoize_value :@commits, Value.new(proc { |c| c.commits(nwo) })
```
parkr: My only concern here is that we limit this to something reasonable like 50 or 100. I believe we may have `auto_paginate` on which would mean if you have 200,000 commits, we'd try to grab all of them. 😳

jerrylopez: Yeaaa... 200,000 commits are definitely not needed. Are you able to set `auto_paginate` to `false` for a specific endpoint?

parkr: We'd have to turn it off for the client while this call is made. Something like: https://github.com/octokit/octokit.rb/issues/554#issuecomment-72337218

jerrylopez: Okay I see what'll need to happen, but unfortunately I'm not proficient in Ruby. How would I go about implementing this?

Issue #93, opened 2017-02-15
Option to disable GitHub metadata when running locally


pacomendes:
We use Jekyll to build documentation locally using the `github-pages` gem. We often don't need Github metadata and the warnings or errors often throw our content writers.

Is it possible to provide a way of disabling Github metadata so that the warnings don't appear on local builds?

benbalter:
> We often don't need Github metadata and the warnings or errors often throw our content writers.

Can you past some of those warnings and errors? We've been working to hide them and fail silently whenever possible.

pacomendes:
Our documentation is on github enterprise and github organization repositories.

We're running locally using bundler with:

Gemfile:
```
source 'https://rubygems.org'
gem 'github-pages', group: :jekyll_plugins
```

Gemfile.lock
```
    ...
    github-pages (122)
      jekyll-github-metadata (= 2.3.1)
   ...
```

_config.yml (to disable error):
`repository: {blank}` or `repository: `username/repo-name`

Jekyll Serve Incremental:
`bundle exec jekyll serve --incremental`

Observed Output with warning:
```
  Server address: http://127.0.0.1:4000/docs/
  Server running... press ctrl-c to stop.
      Regenerating: 1 file(s) changed at 2017-02-16 12:23:24    GitHub Metadata: No GitHub API authentication could be found. Some fields may be missing or have incorrect data.
...done in 19.653869 seconds.
      Regenerating: 1 file(s) changed at 2017-02-16 12:23:44 ...done in 5.33376 seconds.
      Regenerating: 1 file(s) changed at 2017-02-16 12:23:49 ...done in 5.229096 seconds.
      Regenerating: 1 file(s) changed at 2017-02-16 12:23:55 ...done in 6.140592 seconds.
```

The initial regeneration takes long due to the API lookup. Subsequent generations do not have warnings.

We're not using any github metadata for our pages, so would ideally like to disable this lookup to speed up that initial regeneration.

pacomendes:
@benbalter. The delay in regeneration is proving to be an issue for our writers seeing their initial changes locally... Are there any other workarounds that you know of to disable this check/plugin?

pacomendes:
Our work around is just to just the jekyll gem and no longer use github-pages in our Gemfile

`source 'https://rubygems.org'
gem 'jekyll', '3.3.1'`

https://github.com/mmistakes/minimal-mistakes/issues/570

pacomendes:
[...]

PR #95, opened 2017-03-22
Don't double-process the site.github namespace.

parkr:
Should fix https://github.com/jekyll/github-metadata/issues/93#issuecomment-283783238


lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -9,11 +9,17 @@ class GHPMetadataGenerator < Jekyll::Generator
       attr_reader :site
 
       def generate(site)
+        if defined?(@_already_generated)
+          Jekyll::GitHubMetadata.log :debug, "Nothing left to do!"
```
benbalter: We already are exposing users to unnecessary friction with unclear output, I'd suggest making it clear what's going on here.

lib/jekyll-github-metadata/ghp_metadata_generator.rb
```diff
@@ -9,11 +9,17 @@ class GHPMetadataGenerator < Jekyll::Generator
       attr_reader :site
 
       def generate(site)
+        if defined?(@_already_generated)
```
benbalter: Could we hook into `after_init` instead?

benbalter:
Is it possible to tests this?

lib/jekyll-github-metadata/ghp_metadata_generator.rb
parkr: `after_init` was introduced only a few versions ago so this would stop working for older versions. Probably fine, but just calling that out.

In the GitHub Pages gem, we call the configuration reset in `after_reset`, which is called towards the top. Does that seem OK?

lib/jekyll-github-metadata/ghp_metadata_generator.rb
parkr: Agreed – I'm just going to nix it.

parkr:
@benbalter Can you please take another look?

> Is it possible to tests this?

Uhhhh I guess call it twice for the same site and make sure it doesn't call anything?

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -0,0 +1,65 @@
...
+        if site.config["baseurl"].to_s.empty? && !["", "/"].include?(repo.baseurl)
+          site.config["baseurl"] = repo.baseurl
+        end
+      end
+    end
+  end
+end
+
+Jekyll::Hooks.register :site, :after_reset do |site|
+  Jekyll::GitHubMetadata::SiteGitHubMunger.munge(site)
```
benbalter: Would `after_init` save us from having to check if we've already processes it (since it only fires once)?

parkr:
Using `:after_init` now – don't need that test anymore. Thoughts?

parkr:
@jekyllbot: merge

PR #97, opened 2017-06-02
Allow user to set empty baseurl

benbalter:
Via https://github.com/github/pages-gem/issues/350#issuecomment-305796755, it's currently not possible to set a baseurl of "" that the Gem respects. 

https://github.com/jekyll/github-metadata/blob/40a725a6bae050332ad11f491e4af5a61d04ea98/lib/jekyll-github-metadata/site_github_munger.rb#L44 makes it look like it should, but it actually does the exact opposite.

The problem is that [by default, Jekyll sets the baseurl to "", not nil](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/configuration.rb#L50), meaning there's no way to know if it's a user-supplied value or not.

One option would be to respect `site.github.url` or `site.github.baseurl` if passed.

This PR pushes up a failing test and passing tests for the current expected behavior.

If anyone has a better approach, or would like to take a pass at implementing the above, it may be a while before I am able to work on this particular PR.

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -20,6 +20,7 @@ RSpec.describe("integration into a jekyll site") do
     Jekyll::Commands::Build.process({
       "source"      => SOURCE_DIR.to_s,
       "destination" => DEST_DIR.to_s,
+      "gems"        => %w(jekyll-github-metadata),
       "plugins"     => %w(jekyll-github-metadata),
     })
   end
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe("integration into a jekyll site") do
     Jekyll::Commands::Build.process({
       "source"      => SOURCE_DIR.to_s,
       "destination" => DEST_DIR.to_s,
-      "gems"        => %w(jekyll-github-metadata),
+      "plugins"     => %w(jekyll-github-metadata),
     })
   end
   after(:each) do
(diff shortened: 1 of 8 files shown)


parkr:
The route I'd like to take is to change the default in Jekyll from `""` to `nil`. I think that would clear things up here a lot better. We could backport it to v3.4 and ship it with minimal disruption. Would that be 👍 for you?

parkr:
I have a PR here: https://github.com/jekyll/jekyll/pull/6137

parkr:
v3.4.4 is now released with this patch!

marlazinger200:
1

spec/site_github_munger_spec.rb
```diff
@@ -32,6 +32,22 @@
...
+      it "doesn't mangle site.baseurl" do
+        expect(site.config["baseurl"]).to eql("")
+      end
+    end
+
+    context "with site.baseurl set to '/'" do
+      let(:user_config) { { "baseurl" => "/" } }
+
+      it "mangles site.url" do
+        expect(site.config["baseurl"]).to eql("/github-metadata")
```
parkr: You want this case to mangle?

benbalter: That was the original logic. 

* Scenario 1: Users incorrectly set their baseurl as `/`
* Scenario 2: A sophisticated user is proxying their site and wants their baseurl to be `/` despite it being a project page

I'd lean towards mangling, as in scenario 1, the user needs our help more than the user in scenario 2.

benbalter:
@parkr with 3.4.5 out, I believe this is ready for :eyes:.

spec/integration_spec.rb
```diff
@@ -20,7 +20,7 @@ def dest_dir(*files)
     Jekyll::Commands::Build.process({
       "source"      => SOURCE_DIR.to_s,
       "destination" => DEST_DIR.to_s,
-      "gems"        => %w(jekyll-github-metadata),
```
parkr: This only works for Jekyll::VERSION >= 3.5.0.

benbalter:
@jekyllbot: merge

PR #101, opened 2017-07-17
Set default site.title and site.description

benbalter:
For use with things like Jekyll SEO Tag in default templates.


parkr:
If CI passes here, then :shipit:. @jekyllbot can help you out.

parkr:
@jekyllbot: merge

PR #102, opened 2017-07-17
Modernize Travis configuration

parkr:
Update Ruby versions & Jekyll versions.

This should really be automatable.


parkr:
@jekyllbot: merge

PR #104, opened 2017-07-27
Set title and description in dev

benbalter:
A follow up to https://github.com/jekyll/github-metadata/pull/101, the title and description were only being set in production, making local previewing difficult.

This PR updates the logic to set the title and description in dev by moving the conditional statement.

### DIFF
--- a/lib/jekyll-github-metadata.rb
+++ b/lib/jekyll-github-metadata.rb
@@ -18,8 +18,6 @@ module Jekyll
   end
 
   module GitHubMetadata
-    NoRepositoryError = Class.new(Jekyll::Errors::FatalException)
-
     autoload :Client,           "jekyll-github-metadata/client"
     autoload :MetadataDrop,     "jekyll-github-metadata/metadata_drop"
     autoload :Pages,            "jekyll-github-metadata/pages"
@@ -30,13 +28,19 @@ module Jekyll
     autoload :Value,            "jekyll-github-metadata/value"
     autoload :VERSION,          "jekyll-github-metadata/version"
 
+    NoRepositoryError = RepositoryFinder::NoRepositoryError
+
     if Jekyll.const_defined? :Site
       require_relative "jekyll-github-metadata/site_github_munger"
     end
 
     class << self
-      attr_accessor :site
+      attr_reader :repository_finder
       attr_writer :client, :logger
+
+      def site
+        repository_finder.site
+      end
 
       def environment
         Jekyll.respond_to?(:env) ? Jekyll.env : (Pages.env || "development")
@@ -63,13 +67,14 @@ module Jekyll
       end
 
       def repository
-        @repository ||= GitHubMetadata::Repository.new(nwo).tap do |repo|
+        @repository ||= GitHubMetadata::Repository.new(repository_finder.nwo).tap do |repo|
           Jekyll::GitHubMetadata.log :debug, "Generating for #{repo.nwo}"
         end
       end
 
-      def nwo
-        @nwo ||= GitHubMetadata::RepositoryFinder.nwo
+      def site=(new_site)
+        reset!
+        @repository_finder = GitHubMetadata::RepositoryFinder.new(new_site)
       end
 
       def reset!
--- a/lib/jekyll-github-metadata.rb
+++ b/lib/jekyll-github-metadata.rb
@@ -24,6 +24,7 @@ module Jekyll
     autoload :MetadataDrop,     "jekyll-github-metadata/metadata_drop"
     autoload :Pages,            "jekyll-github-metadata/pages"
     autoload :Repository,       "jekyll-github-metadata/repository"
+    autoload :RepositoryFinder, "jekyll-github-metadata/repository_finder"
     autoload :RepositoryCompat, "jekyll-github-metadata/repository_compat"
     autoload :Sanitizer,        "jekyll-github-metadata/sanitizer"
     autoload :Value,            "jekyll-github-metadata/value"
@@ -34,7 +35,7 @@ module Jekyll
     end
 
     class << self
-      attr_accessor :repository
+      attr_accessor :site
       attr_writer :client, :logger
 
       def environment
@@ -61,9 +62,18 @@ module Jekyll
         @client ||= Client.new
       end
 
+      def repository
+        @repository ||= GitHubMetadata::Repository.new(nwo).tap do |repo|
+          Jekyll::GitHubMetadata.log :debug, "Generating for #{repo.nwo}"
+        end
+      end
+
+      def nwo
+        @nwo ||= GitHubMetadata::RepositoryFinder.nwo
+      end
+
       def reset!
-        @logger = nil
-        @client = nil
+        @logger = @client = @repository = @nwo = @site = nil
       end
     end
   end
--- a/lib/jekyll-github-metadata/site_github_munger.rb
+++ b/lib/jekyll-github-metadata/site_github_munger.rb
@@ -12,6 +12,7 @@ module Jekyll
 
       def munge!
         Jekyll::GitHubMetadata.log :debug, "Initializing..."
+        Jekyll::GitHubMetadata.repository = repository
 
         # This is the good stuff.
         site.config["github"] = github_namespace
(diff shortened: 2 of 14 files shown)


lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -16,9 +16,8 @@ def munge!
         # This is the good stuff.
         site.config["github"] = github_namespace
 
-        return unless should_add_fallbacks?
-        add_url_and_baseurl_fallbacks!
         add_title_and_description_fallbacks!
```
parkr: I wonder if we should bail early if:

1. No auth token is present (if they rebuild frequently, they'll start getting 403 Rate Limit Exceeded errors which are confusing)
2. There is no internet connectivity

spec/site_github_munger_spec.rb
```diff
@@ -83,6 +84,10 @@
     before(:each) do
       Jekyll::GitHubMetadata.client = Jekyll::GitHubMetadata::Client.new({ :access_token => "" })
     end
+    before do
```
parkr: Why not `before(:each)`? IIRC `before` is discouraged because of the random ordering of each test.

benbalter: Not related to this PR, so I'll pull it out (debugging artifact).

lib/jekyll-github-metadata/site_github_munger.rb
benbalter: :+1: Both are a new feature in themselves, and out of scope for this PR, but 👍 to tackling those in a subsequent pass.

spec/test-site/index.html
```diff
@@ -9,7 +9,7 @@
 </head>
 <body>
     <div id="output"></div>
-    <script src="renderjson.js" type="text/javascript" charset="utf-8"></script>
+    <script src="{{ "renderjson.js" | absolute_url }}" type="text/javascript" charset="utf-8"></script>
```
benbalter: Added this to ensure the internals that use `url` and `baseurl` don't blow up when either is a Proc.

benbalter:
Also, to memorialize things here publicly, chatting with @parkr, I didn't like the idea that we'd be making API calls for every build, even if the user doesn't use `site.title` or `site.description`. By moving the calls to Value wrapped Procs, in theory, a site that has a baseurl set and doesn't call `site.title`, `site.description`, `site.url` or `site.github.*` should not make an API call.

parkr:
Thanks!

If you give the proc an arity of 2, it'll pass you the right repo too! First argument is the Client, second is the Repository :smile:

benbalter:
@parkr this could use a second set of :eyes: now, but otherwise should be good to go. Per our discussion:

* Created RepositoryFinder to centralize `nwo` logic
* Centralized and memoized `site` and `repository` on the GitHubMetadata class
* Added unit tests for each expected key/value in MetadataDrop 
* Added a test to ensure we're testing all values in MetadataDrop
* Value now forwards `+`, and `eql?` to the value to avoid `to_s` calls

As a result, when called locally, by default, the plugin should make no external API calls. Only if, for example, `site.title` is called, and is not defined, or `site.github.*` is called, will it make an API call.

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -4,10 +4,12 @@
 module Jekyll
   module GitHubMetadata
     class SiteGitHubMunger
-      attr_reader :site
+      extend Forwardable
+
+      def_delegators :"Jekyll::GitHubMetadata", :site, :repository
```
parkr: Do these need to be public?

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -35,13 +36,14 @@ def github_namespace
       end
 
       def drop
-        @drop ||= MetadataDrop.new(site)
+        @drop ||= MetadataDrop.new
```
parkr: This makes me feel woozy. I wonder if keeping the `site` on the global GitHubMetadata is a bad idea? I'm ok with `client` and `repository` at that level, but the whole `site` object scares me.

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -35,13 +36,14 @@ def github_namespace
...
       end
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
-        site.config["url"] ||= repository.url_without_path
-        site.config["baseurl"] = repository.baseurl if should_set_baseurl?
+        site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })
```
parkr: 😍

lib/jekyll-github-metadata/value.rb
```diff
@@ -3,6 +3,9 @@
 module Jekyll
   module GitHubMetadata
     class Value
+      extend Forwardable
+      def_delegators :render, :+, :to_s, :to_json, :eql?
```
parkr: Clever. Can you memoize `render` so it doesn't render on each call? Or use `:value` instead of `:render`.

I'd also like to have a test for each of these – it can be really basic, but I just want to ensure they don't break.

Why `#+`?

Per [`Object#hash`](https://ruby-doc.org/core-2.4.1/Object.html#method-i-hash), if you specify `#eql?` you must *also* specify `#hash`:

> Generates an Integer hash value for this object. This function must have the property that `a.eql?(b)` implies `a.hash == b.hash`.

lib/jekyll-github-metadata/value.rb
```diff
@@ -39,14 +42,6 @@ def render
         raise e
       end
 
-      def to_s
-        render.to_s
-      end
-
-      def to_json(*)
```
parkr: Unfortunately, we can't just alias this method to `#render`. `to_json` must be able to accept a `state` parameter. We [do that in Jekyll::Drops::Drop](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/drops/drop.rb#L153-L161) to great effect. (See how `state` affects things in [`DocumentDrop#hash_for_json`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/drops/document_drop.rb#L40-L53).)

```ruby
def to_json(state = nil)
  require "json"
  JSON.generate(value, state)
end
```

benbalter: Because it's `def_delegators` (plural), I believe it will delegate `Value#to_json` to calling `to_json` on the result of `#render`, meaning it should accept the second argument so long as the result of `#render` does.

lib/jekyll-github-metadata/site_github_munger.rb
benbalter: `site` was already public, but made repo private via 66f4b4a.

lib/jekyll-github-metadata.rb
```diff
@@ -61,9 +62,18 @@ def client
...
       end
 
+      def repository
+        @repository ||= GitHubMetadata::Repository.new(nwo).tap do |repo|
+          Jekyll::GitHubMetadata.log :debug, "Generating for #{repo.nwo}"
+        end
+      end
+
+      def nwo
+        @nwo ||= GitHubMetadata::RepositoryFinder.nwo
```
benbalter: If we did that, JekyllGitHubMetadata would have to know (or be passed) the site variable (which you seemed to indicate the opposite below).

lib/jekyll-github-metadata/site_github_munger.rb
benbalter: Thoughts on a better solution? We need site in order to build repository.

lib/jekyll-github-metadata/value.rb
benbalter: Memoized via 7027532.

> Why `#+`?

Workaround so we don't need to backport https://github.com/jekyll/jekyll/pull/6253.

> Per Object#hash, if you specify #eql? you must also specify #hash:

Done via f0662b0.

benbalter: > I'd also like to have a test for each of these – it can be really basic, but I just want to ensure they don't break.

added via 4ea6404

lib/jekyll-github-metadata.rb
parkr: My struggle is that it's too easy for this data to get all out of sync. I want the behavior that if `GitHubMetadata.site` changes, that `GitHubMetadata.repository` changes with it.

```ruby
module GitHubMetadata
  def self.site
    repository_finder.site
  end

  def self.site=(new_site)
    reset!
    @repository_finder = GitHubMetadata::RepositoryFinder.new(new_site)
  end

  def self.repository
    @repository ||= GitHubMetadata::Repository.new(repository_finder.nwo).tap do |repo|
      Jekyll::GitHubMetadata.log :debug, "Generating for #{repo.nwo}"
    end
  end

  private
  attr_reader :repository_finder
    
  end
end
```

This enforces a safe top-level environment: when `site` is set to some other value, it resets everything and creates a new `repository_finder` instance with the new site.

parkr:
@jekyllbot: merge +minor

PR #105, opened 2017-07-27
Expose repo visibility

benbalter:
This PR adds a `private` key to the `site.github` namespace. The value will be `true` if the repo is private, otherwise, it will be `false`. This will allow a site to know it's visibility (for example, to conditionally expose an edit button via a theme).


benbalter:
@jekyllbot: merge +minor

PR #106, opened 2017-08-09
Expose repo license

benbalter:
A follow up to https://github.com/jekyll/github-metadata/pull/105, this PR exposes the repository's licenses as `site.github.license`. I'd like to use this to conditionally expose an "This site is open source, help improve this page" link if a repo is public and has an open source license (versus public but not necessarily looking for contributions).

In doing this, I realized that we had a handful of very similar methods on the Repository model that simply delegated to a hash. Forwardable can't delegate to `[]`, only to the method itself, so I created a quick `def_hash_delegator` helper, to allow us to DRY things up a bit and delegate a method more easily to a hash key. It works exactly like `def_delegator`, e.g. `def_hash_delegator :repo_info, :license` is the same as:

```ruby
def license
  repo_info["license"]
end
```

and `def_hash_delegator :repo_info, :description, :tagline` is the same as:

```ruby
def tagline
  repo_info["description"]
end
```


lib/jekyll-github-metadata/repository.rb
```diff
@@ -2,6 +2,33 @@ module Jekyll
...
+      # Defines an instance method that delegates to a hash's key
+      #
+      # hash   - a symbol representing the instance method to delegate to. The
+      #          instance method should return a hash or respond to #[]
+      # key    - the key to call within the hash
+      # method - (optional) the instance method the key should be aliased to.
+      #          If not specified, defaults to the hash key
+      #
+      # Returns a symbol representing the instance method
+      def self.def_hash_delegator(hash, key, method = nil)
```
parkr: The order of these is hard to read. I think this is easier to read:

```
def self.def_hash_delegator(method, hash, key)
```

Having the method name only some of the times is hard to grok. 

I guess matching it with `def_delegator` is a noble goal. In that case, I'd say we should *always* require the `method` argument. Also, for 👀 sake, maybe align each argument in a table-like format for easy scanning?

benbalter: Good call. Implemented via 211fe3d.

benbalter:
@jekyllbot: merge +minor

PR #108, opened 2017-08-17
GitHub edit link tag

benbalter:
This PR exposes a `github_edit_link` tag that generates links to edit the current page on GitHub.

### To generate a link

```liquid
<p>This site is open source. {% github_edit_link "Improve this page" %}</p>
```

Produces:

```html
<p>This site is open source. <a href="https://github.com/benbalter/jekyll-edit-link/edit/master/README.md">Improve this page</a></p>
```

### To generate a path

If you'd prefer to build your own link, simply don't pass link text

```liquid
<p>This site is open source. <a href="{% github_edit_link %}">Improve this page</a></p>
```

Produces:

```html
<p>This site is open source. <a href="https://github.com/benbalter/jekyll-edit-link/edit/master/README.md">Improve this page</a></p>
```

In order to add the tag, I made to related changes:

* Broke the various spec helpers into `spec/spec_helpers/foo_helper.rb` to keep `spec/spec_helpers.rb` more managable

* Broke the README into individual documents, and moved to the `docs/` folder since adding the edit tag docs made the README a bit unwieldy.


docs/authentication.md
```diff
@@ -0,0 +1,31 @@
...
+
+These tokens are easy to use and delete so if you move around from machine-to-machine, we'd recommend this route. Set `JEKYLL_GITHUB_TOKEN` to your access token (with `public_repo` scope) when you run `jekyll`, like this:
+
+```bash
+$ JEKYLL_GITHUB_TOKEN=123abc [bundle exec] jekyll serve
+```
+
+### 2. `~/.netrc`
+
+If you prefer to use the good ol' `~/.netrc` file, just make sure the `netrc` gem is bundled and run `jekyll` like normal. So if I were to add it, I'd add `gem 'netrc'` to my `Gemfile`, run `bundle install`, then run `bundle exec jekyll build`. The `machine` directive should be `api.github.com`.
```
parkr: Maybe I should include an example...

```text
machine api.github.com
    login github-username
    password 123abc-your-token
```

spec/spec_helpers/env_helper.rb
```diff
@@ -0,0 +1,29 @@
+
```
parkr: empty space!

spec/spec_helpers/fixture_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+  def config_defaults
+    {
+      "source"      => source_dir.to_s,
+      "destination" => dest_dir.to_s,
+      "plugins"     => ["jekyll-github-metadata"],
+      "gems"        => ["jekyll-github-metadata"],
+    }
+  end
+
+  def make_page(options = {})
```
parkr: why not call this `data` or `page_data` if it's only every being used as `page.data`?

spec/spec_helpers/fixture_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    page = Jekyll::Page.new site, config_defaults["source"], "", "page.md"
```
parkr: parens around the 4 arguments to `.new` would help readability here 👀

spec/spec_helpers/fixture_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    page.data = options
```
parkr: you could use .tap

```text
Jekyll::Page.new(site, config_defaults["source"], "", "page.md").tap { |page| page.data = data }
```

returns the Page object

lib/jekyll-github-metadata/edit-link-tag.rb
```diff
@@ -0,0 +1,102 @@
...
+      #                instance method should return a hash or respond to #[]
+      # key         - the key to call within the hash
+      # method      - (optional) the instance method the key should be aliased to.
+      #               If not specified, defaults to the hash key
+      # default     - (optional) value to return if value is nil (defaults to nil)
+      #
+      # Returns a symbol representing the instance method
+      def self.def_hash_delegator(hash_method, key, method, default = nil)
+        define_method(method) do
+          hash = send(hash_method)
```
benbalter: If we did, we'd have to make `site`, `site_github`, `source`, and `page` public.

lib/jekyll-github-metadata/edit-link-tag.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def parts
+        @parts ||= [repository_url, "edit/", branch, source_path, page_path]
+      end
+
+      def parts_normalized
+        @parts_normalized ||= parts.map.with_index do |part, index|
+          part = remove_leading_slash(part.to_s)
+          part = ensure_trailing_slash(part) unless index == parts.length - 1
```
benbalter: If the url is `https://github.com/foo/bar/edit/master/page.md`, we don't want the last element `page.md` to be given a trailing slash (and can't rely on the extension, because the repo might be, `foo.js`).

spec/spec_helpers/fixture_helper.rb
benbalter: It's used both in the fixture helper and in the edit link tag spec to create the context, which needs the page object, not the page data.

benbalter:
@parkr thanks for the feedback. I believe I've implemented your suggestions and answered your questions to the best of my ability. Mind taking another look? :smile:

spec/spec_helpers/fixture_helper.rb
parkr: Ah, I meant the argument. `s/options/data/`. Sorry for the confusion!

lib/jekyll-github-metadata/edit-link-tag.rb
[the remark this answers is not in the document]
parkr: Could you leave a comment in the code to that effect or give it a tiny helper method which explains that?

benbalter:
@jekyllbot: merge

DirtyF:
Is it supposed to work locally? 

My default branch is `master`, my `_config.yml`also has `branch:master`and here is the link that gets generated when testing on my machine:

`https://github.com/DirtyF/frank.taillandier.me/edit/gh-pages/_posts/2017-05-21-sept-ans-de-sud-web.md``

benbalter:
@DirtyF I believe the following should work in your config:

```yml
github:
  source:
    branch: master
```

DirtyF:
I hoped so but something is wrong:

GitHub API is returning `master`

```sh
$ http --body https://api.github.com/repos/DirtyF/frank.taillandier.me/branches
[
    {
        "commit": {
            "sha": "15a70597f9aee29fcb4c7ae354489b0b560230e8",
            "url": "https://api.github.com/repos/DirtyF/frank.taillandier.me/commits/15a70597f9aee29fcb4c7ae354489b0b560230e8"
        },
        "name": "master"
    }
]
```

Whether `{{ site.github.source.branch }}` is returning `gh-pages`. 😲

benbalter:
The tag uses `site.github.source.branch` which comes from https://github.com/jekyll/github-metadata/blob/master/lib/jekyll-github-metadata/metadata_drop.rb#L61, which itself is delegated to  https://github.com/jekyll/github-metadata/blob/master/lib/jekyll-github-metadata/repository.rb#L125-L127 which will either be the source as returned from the *pages* endpoint, or if it can't hit that endpoint, [`gh-pages` for user pages](https://github.com/jekyll/github-metadata/blob/master/lib/jekyll-github-metadata/repository_compat.rb#L58-L63).

If `site.github` is a hash, [we merge the drop (the logic above) with the user-supplied value](https://github.com/jekyll/github-metadata/blob/master/lib/jekyll-github-metadata/site_github_munger.rb#L32-L33). It could be that Jekyll's `deep_merge_hashes` isn't recursive enough, but in theory, if you supply `site.github.source.branch` it should prefer that over the calculated value.

DirtyF:
Thanks for taking the time to explain how this works under the hood. 🔧 

> in theory, if you supply `site.github.source.branch` it should prefer that over the calculated value.

Well, as I said, In my case, it keeps the default value 'gh-pages`.

If this only supposed to work with repos that have GitHub Pages activated, I guess it's an edge case and out of scope.

benbalter:
@DirtyF You found a 🐛 🐞! Fix in progress via https://github.com/jekyll/github-metadata/pull/110.

PR #109, opened 2017-08-25
define path with __dir__

DirtyF:
/cc @bogdanvlviv


benbalter:
@jekyllbot: merge

benbalter:
Thanks @DirtyF!

PR #110, opened 2017-08-29
Allow user values to override drop-determined values

benbalter:
Via https://github.com/jekyll/github-metadata/pull/108#issuecomment-325817689, it appears user-supplied `site.github` values are being overwritten by drop-supplied values.

It appears in https://github.com/jekyll/github-metadata/commit/54942a488f493a346d709c208799558afc012a5a#diff-165b64b2f6bce98447b6c550f3205ad9L26, the order of the arguements passed to `Jekyll::Utils.deep_merge_hashes` were reversed since [the right overrides the left value](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/utils.rb#L39-L54).

This PR adds failing tests via a7d2b4b to ensure `site.github` is set to the expected value given different user-supplied inputs for `site.github` with d36d1af reversing the order of the arguments.

/cc @DirtyF

### DIFF
--- a/lib/jekyll-github-metadata/site_github_munger.rb
+++ b/lib/jekyll-github-metadata/site_github_munger.rb
@@ -30,7 +30,7 @@ module Jekyll
         when nil
           drop
         when Hash
-          Jekyll::Utils.deep_merge_hashes(site.config["github"], drop)
+          Jekyll::Utils.deep_merge_hashes(drop, site.config["github"])
         else
           site.config["github"]
         end
--- a/lib/jekyll-github-metadata/site_github_munger.rb
+++ b/lib/jekyll-github-metadata/site_github_munger.rb
@@ -66,5 +66,5 @@ module Jekyll
 end
 
 Jekyll::Hooks.register :site, :after_init do |site|
-  Jekyll::GitHubMetadata::SiteGitHubMunger.new(site).munge! unless Jekyll.env == "test"
+  Jekyll::GitHubMetadata::SiteGitHubMunger.new(site).munge!
 end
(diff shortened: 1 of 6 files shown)


DirtyF:
@benbalter FWIW I just did [a test with this branch](https://github.com/DirtyF/frank.taillandier.me/commit/fa138a9b5242791dad32d8dc64cf5dc82715527c) and it does still output the drop value from GitHub, not the one from the local config file, even with the switch of the arguments of `deep_merge_hashes`. 🤔

benbalter:
@DirtyF backported (forward-ported?) https://github.com/jekyll/jekyll/pull/6338 via 7a97ed0. Let me know if things work as expected now?

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -30,7 +30,7 @@ def github_namespace
         when nil
           drop
         when Hash
-          Jekyll::Utils.deep_merge_hashes(site.config["github"], drop)
+          Jekyll::Utils.deep_merge_hashes(drop, site.config["github"])
```
parkr: Flipping this may have some unintended consequences (because people expect the opposite) so we should be prepared for upset users.

DirtyF:
@benbalter  the issue is not resolved in my case`{{ site.github.source.branch }}` still output `gh-pages`  😞

benbalter:
@DirtyF see https://github.com/jekyll/jekyll/pull/6338#issuecomment-327234319. Mind giving it one more try? (I tested locally with a test site and it worked as expected).

benbalter:
> Test: DirtyF/frank.taillandier.me@85987a8

Tested locally and go the following (expected) output:

```html
  <footer>
    <p>Une correction à apporter ? <a href="https://github.com//edit/master/_posts/2017-08-27-securite-psychologique.md">Éditez cet article</a>.</p>
  </footer>
```

benbalter:
@jekyllbot: merge +minor.

PR #113, opened 2017-09-11
Warn and do nothing when site.name is set

benbalter:
Per https://github.com/jekyll/jekyll-seo-tag/issues/238#issuecomment-328594374.

### DIFF
--- a/lib/jekyll-github-metadata/site_github_munger.rb
+++ b/lib/jekyll-github-metadata/site_github_munger.rb
@@ -47,19 +47,30 @@ module Jekyll
         site.config["baseurl"] = Value.new("baseurl", proc { |_c, r| r.baseurl })
       end
 
+      def add_title_and_description_fallbacks!
+        if should_warn_about_site_name?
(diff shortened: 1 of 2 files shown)


lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -47,20 +47,31 @@ def add_url_and_baseurl_fallbacks!
...
       # Baseurls should never be "/". See http://bit.ly/2s1Srid
       def should_set_baseurl?
         site.config["baseurl"].nil? || site.config["baseurl"] == "/"
       end
 
-      def add_title_and_description_fallbacks!
-        site.config["title"] ||= Value.new("title", proc { |_c, r| r.name })
-        site.config["description"] ||= Value.new("description", proc { |_c, r| r.tagline })
-      end
-
```
benbalter: Moved above to group the bang and should methods together.

benbalter:
@jekyllbot: merge +minor.

PR #115, opened 2017-11-11
Docs: use plugins config key

Pilabio:


pathawks:
Looks great. Thanks @Pilabio :tada::beers:

@jekyllbot: merge +docs

PR #128, opened 2018-06-03
update GitHub Pages domain suffix to github.io

inntran:
Since GitHub has moved from .com to .io to host gh-pages, https://blog.github.com/2013-04-05-new-github-pages-domain-github-io/
It makes sense to update here as well.

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -147,7 +147,7 @@ module Jekyll
 
       def user_page_domains
         domains = [default_user_domain]
-        domains.push "#{owner}.github.com".downcase unless Pages.enterprise?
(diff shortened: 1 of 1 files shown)


inntran:
I checked the failed test, run script/cibuild locally on macOS 10.13, with Ruby 2.5.1, no errors.

parkr:
This is a legacy fallback! Check the value of that array. It's all _possible_ domains. For non-Enterprise users, that includes `username.github.com`. The implementation for `default_user_domain` will be `username.github.io`, but we fall check `username.github.com` too so we don't break compatibility.

inntran:
I understand. The reason I thought it was an issue: on local setup, it's recognized as a project_page, but on GitHub.com it's correctly recognized as a user_page, showing different links(project vs user/org) in <header>.

PR #134, opened 2018-09-10
Add Appveyor config

DirtyF:
- Add continuous integration on Windows platform.
- Test against two latest minor versions of Jekyll
- Bump Rubocop to 0.57 (latest version to target Ruby 2.1)

/cc @jekyll/windows

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,30 @@
+version: "{build}"
+
+clone_depth: 10
+
+branches:
+  only:
+    - master
+
+build: off
+
+install:
+  - SET PATH=C:\Ruby%RUBY_FOLDER_VER%\bin;%PATH%
+  - bundle install --retry 5 --jobs=%NUMBER_OF_PROCESSORS% --clean --path vendor\bundle
+
+environment:
+  matrix:
+    - RUBY_FOLDER_VER: "25"
+    - RUBY_FOLDER_VER: "25-x64"
+    - RUBY_FOLDER_VER: "24"
+    - RUBY_FOLDER_VER: "23"
+
+test_script:
+  - ruby --version
+  - gem --version
+  - bundler --version
+  - bash ./script/cibuild
+
+cache:
+  # If one of the files after the right arrow changes, cache will be skipped
+  - 'vendor\bundle -> appveyor.yml,Gemfile,jekyll-sass-converter.gemspec'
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -0,0 +1,30 @@
+version: "{build}"
+
+clone_depth: 10
+
+branches:
+  only:
+    - master
```
XhmikosR: Are you sure about only master? Why not test all branches?

appveyor.yml
```diff
@@ -0,0 +1,30 @@
...
+
+test_script:
+  - ruby --version
+  - gem --version
+  - bundler --version
+  - bash ./script/cibuild
+
+cache:
+  # If one of the files after the right arrow changes, cache will be skipped
+  - 'vendor\bundle -> appveyor.yml,Gemfile,jekyll-sass-converter.gemspec'
```
XhmikosR: Oops, here it seems there's a typo @DirtyF

appveyor.yml
```diff
@@ -0,0 +1,30 @@
...
+
+test_script:
+  - ruby --version
+  - gem --version
+  - bundler --version
+  - bash ./script/cibuild
+
+cache:
+  # If one of the files after the right arrow changes, cache will be skipped
+  - 'vendor\bundle -> appveyor.yml,Gemfile,jekyll-github-metadata.gemspec'
```
XhmikosR: BTW, IIRC we can use a wildcard here. Something to try in the future.

jekyll-github-metadata.gemspec
```diff
@@ -23,5 +23,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "netrc"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "rspec"
-  spec.add_development_dependency "rubocop", "0.51"
+  spec.add_development_dependency "rubocop", "~> 0.59.0"
```
parkr: Should we be using our version of rubocop? Or for another PR?

DirtyF: As Jekyll 3.8.3 still target 2.1, we can lock to 0.57 for now, I'll update this PR.

ashmaroli:
@DirtyF Since you're not dropping support for Ruby 2.2 and 2.1, I suggest that you undo all changes to Ruby files and run a fresh session of `rubocop -a` with `v0.57.x`, mainly because the string literals need to be frozen explicitly for older Ruby versions..

DirtyF:
@ashmaroli I did, but looks like it doesn't revert all the changes made by 0.59 autocorrect, I'll give it another pass.

DirtyF:
@jekyllbot: merge +dev

PR #136, opened 2018-10-21
Fixes for repository detection on Windows

GICodeWarrior:
The repository detection code wasn't portable to Windows.

These changes fix detection on Windows and maintain portability with other platforms.

### DIFF
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -48,7 +48,7 @@ module Jekyll
       end
 
       def git_exe_path
-        exts = (ENV['PATHEXT'] || '').split(File::PATH_SEPARATOR)
+        exts = (ENV['PATHEXT'] || '').split(File::PATH_SEPARATOR).push('')
         cmds = exts.map { |ext| "git#{ext}" }
         ENV["PATH"].to_s
           .split(File::PATH_SEPARATOR)
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -46,13 +46,11 @@ module Jekyll
       end
 
       def git_remotes
-        output = ""
-        begin
-          _process, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
-        rescue Errno::ENOENT => e
-          Jekyll.logger.warn "Not Installed:", e.message
-        end
+        _process, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
         output.to_s.strip.split("\n")
+      rescue Errno::ENOENT => e
+        Jekyll.logger.warn "Not Installed:", e.message
+        ""
       end
 
       def git_remote_url
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -47,19 +47,13 @@ module Jekyll
         repo if repo && repo.is_a?(String) && repo.include?("/")
       end
 
-      def git_exe_path
-        exts = (ENV["PATHEXT"] || "").split(File::PATH_SEPARATOR).push("")
-        cmds = exts.map { |ext| "git#{ext}" }
-        ENV["PATH"].to_s
-          .split(File::PATH_SEPARATOR)
-          .map { |path| cmds.map { |cmd| File.join(path, cmd) } }
-          .flatten
-          .find { |path| File.executable?(path) && !File.directory?(path) }
-      end
-
       def git_remotes
-        return [] if git_exe_path.nil?
-        output, _status = Open3.capture2(git_exe_path, "remote", "--verbose")
+        output = ""
+        begin
+          _p, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
+        rescue Errno::ENOENT => e
+          Jekyll.logger.warn "Git is not installed: ", e.message
+        end
         output.to_s.strip.split("\n")
       end
 
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -50,7 +50,7 @@ module Jekyll
         begin
           _process, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
         rescue Errno::ENOENT => e
-          Jekyll.logger.warn "Git is not installed: ", e.message
+          Jekyll.logger.warn "Not Installed:", e.message
         end
         output.to_s.strip.split("\n")
       end
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'open3'
+require "open3"
 
 module Jekyll
   module GitHubMetadata
@@ -48,7 +48,7 @@ module Jekyll
       end
 
       def git_exe_path
-        exts = (ENV['PATHEXT'] || '').split(File::PATH_SEPARATOR).push('')
+        exts = (ENV["PATHEXT"] || "").split(File::PATH_SEPARATOR).push("")
         cmds = exts.map { |ext| "git#{ext}" }
         ENV["PATH"].to_s
           .split(File::PATH_SEPARATOR)
@@ -59,7 +59,7 @@ module Jekyll
 
       def git_remotes
         return [] if git_exe_path.nil?
-        output, status = Open3.capture2(git_exe_path, 'remote', '--verbose')
+        output, _status = Open3.capture2(git_exe_path, "remote", "--verbose")
         output.to_s.strip.split("\n")
       end
 
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -46,10 +46,13 @@ module Jekyll
       end
 
       def git_exe_path
+        exts = (ENV['PATHEXT'] || '').split(File::PATH_SEPARATOR)
+        cmds = exts.map { |ext| "git#{ext}" }
         ENV["PATH"].to_s
           .split(File::PATH_SEPARATOR)
-          .map { |path| File.join(path, "git") }
-          .find { |path| File.exist?(path) }
+          .map { |path| cmds.map { |cmd| File.join(path, cmd) } }
+          .flatten
+          .find { |path| File.executable?(path) && !File.directory?(path) }
       end
 
       def git_remotes
--- a/lib/jekyll-github-metadata/repository_finder.rb
+++ b/lib/jekyll-github-metadata/repository_finder.rb
@@ -50,7 +50,7 @@ module Jekyll
         output.to_s.strip.split("\n")
       rescue Errno::ENOENT => e
         Jekyll.logger.warn "Not Installed:", e.message
-        ""
+        []
       end
 
       def git_remote_url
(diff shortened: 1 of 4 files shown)


ashmaroli:
> The repository detection code wasn't portable to Windows.

@GICodeWarrior Can you elaborate on how exactly you came across this issue..?
I'd like to reproduce it at my end before reviewing this proposal..

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -46,15 +48,19 @@ def nwo_from_config
       end
 
       def git_exe_path
+        exts = (ENV["PATHEXT"] || "").split(File::PATH_SEPARATOR).push("")
```
benbalter: What's going on here?

GICodeWarrior: `PATHEXT` on Windows is a list of file extensions respected by cmd.exe.
http://environmentvariables.org/PathExt

This code splits PATHEXT into an array of extensions and adds a blank extension to the end (for odd Windows setups and normal non-Windows setups).

```
irb(main):001:0> (ENV["PATHEXT"] || "").split(File::PATH_SEPARATOR).push("")
=> [".COM", ".EXE", ".BAT", ".CMD", ".VBS", ".VBE", ".JS", ".JSE", ".WSF", ".WSH", ".MSC", ".RB", ".RBW", ""]
```

On non-Windows platforms `exts` is only `[""]`.

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -46,15 +48,19 @@ def nwo_from_config
...
+        cmds = exts.map { |ext| "git#{ext}" }
```
ashmaroli: Is there any known scenario where `git.exe` is **not the Git Executable** on Windows? If not, lets not waste time searching for those paths and zero in on `~/git.exe` straight away.

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -46,15 +48,19 @@ def nwo_from_config
...
         ENV["PATH"].to_s
           .split(File::PATH_SEPARATOR)
-          .map { |path| File.join(path, "git") }
-          .find { |path| File.exist?(path) }
+          .map { |path| cmds.map { |cmd| File.join(path, cmd) } }
+          .flatten
+          .find { |path| File.executable?(path) && !File.directory?(path) }
```
ashmaroli: Is this branching really necessary?
When is an executable ever a directory?

ashmaroli:
@GICodeWarrior Okay, now that I have the proper context to this proposal, I understand the motivation behind this PR.
But my opinion on the changes proposed is that it all appears to be *too hacky*!

I'd personally recommend abstracting the entire "Windows support logic" into a "separate set" (or class) of methods.

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -46,15 +48,19 @@ def nwo_from_config
...
       end
 
       def git_remotes
         return [] if git_exe_path.nil?
-        `#{git_exe_path} remote --verbose`.to_s.strip.split("\n")
+        output, _status = Open3.capture2(git_exe_path, "remote", "--verbose")
```
ashmaroli: @parkr @benbalter  Can we drop support for Jekyll versions `< 3.4.0` and use `Jekyll::Utils::Exec.run` instead of using `Open3` directly..?

DirtyF: Good idea, as GitHub pages uses currently `3.7.4` it shouldn't be a problem.

lib/jekyll-github-metadata/repository_finder.rb
GICodeWarrior: I don't really have visibility into how people have their machines setup, but having a `git.bat` file to wrap Git is one edge case I can think of.

Random example: https://gist.github.com/duebbert/c4f26c9d0691b7d4a0e932e7d9f36271

Technically all of the extensions in `PATH_EXT` can be used plainly as `git` from the command line on Windows if someone added a wrapper or alternate implementation to their `PATH`.

If you want to simplify it, I recommend skipping the `PATH` search altogether and allowing the `Open3`/`Exec.run` call to outsource that to the OS.

Do you know why this `PATH` code exists in the first place?

lib/jekyll-github-metadata/repository_finder.rb
GICodeWarrior: Directories are always executable.  So if there is a `git` directory in `PATH`, we don't want to select it.

Windows:
```
irb(main):006:0> File.directory?('/Users')
=> true
irb(main):007:0> File.executable?('/Users')
=> true
```

Linux:
```
irb(main):002:0> File.directory?('/home')
=> true
irb(main):003:0> File.executable?('/home')
=> true
```

GICodeWarrior:
If I get back to this, I'll update with a version that removes all the `PATH`ing code.  We can defer to the OS for the correct logic and simply rescue the exception when Git is not available.

lib/jekyll-github-metadata/repository_finder.rb
ashmaroli: Delegating to Windows can be bothersome IMO.
Some machines respond to `where` while some responds to `where.exe` while the others do not have either of the two programs in their `PATH`..

ashmaroli:
@GICodeWarrior Delegating to the OS is probably not a good solution here.. (can not be reliably cross-platform)..
Please proceed by incorporating `PATHEXT` with the rest of the suggestions I've listed above..

GICodeWarrior:
Can you clarify how allowing the platform to leverage it's own PATH mechanisms will be less reliable than a bespoke implementation of those mechanisms in this library?

GICodeWarrior:
Here are some examples of how Jekyll uses `Jekyll::Utils::Exec.run`: 

https://github.com/jekyll/jekyll/blob/0728ccf08b08759d75b6bac40091151a80556565/features/support/helpers.rb#L88-L119

ashmaroli:
> Can you clarify how allowing the platform to leverage it's own PATH mechanisms will be less reliable

How would a developer on UNIX-based systems go about this..?
Their system *would probably have the `which` program installed..*
```bash
$ which git
# => /path/to/git
```
And how would a developer on Windows go about it?
They *may* or *may not* have a `which.exe` program installed.
Though, there's a chance that their system *probably has `where.exe` program installed*
Even then, the invocation is different based on the shell used:

```sh
# on default Windows Command Prompt

where git
# => C:\path\to\git.exe
```
```sh
# on Windows PowerShell

where.exe git
# => C:\path\to\git.exe
```

ashmaroli:
Well, it's a different story entirely if you're no longer intending to find the path to `git` executable at all..

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -45,16 +45,14 @@ def nwo_from_config
...
-          .map { |path| File.join(path, "git") }
-          .find { |path| File.exist?(path) }
-      end
-
       def git_remotes
-        return [] if git_exe_path.nil?
-        `#{git_exe_path} remote --verbose`.to_s.strip.split("\n")
+        output = ""
+        begin
+          _p, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
```
ashmaroli: we prefer verbose variable names where possible.
```suggestion
          _process, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
```

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -45,16 +45,14 @@ def nwo_from_config
...
+        rescue Errno::ENOENT => e
+          Jekyll.logger.warn "Git is not installed: ", e.message
```
ashmaroli: The first argument to Jekyll.logger should *ideally be 20 chars or lesser* to avoid misalignment
```suggestion
          Jekyll.logger.warn "Not Installed:", e.message
```

ashmaroli:
@GICodeWarrior One other thing.. We need to drop support for Jekyll versions older than `v3.4.0` since the utility method doesn't exist in those versions

lib/jekyll-github-metadata/repository_finder.rb
ashmaroli: Today I Learned <sup>TM</sup> :+1:

ashmaroli:
@parkr @benbalter Should this warrant a bump in the major version since we're dropping support for Jekyll versions older than `v3.4.0`..?

lib/jekyll-github-metadata/repository_finder.rb
```diff
@@ -45,16 +45,14 @@ def nwo_from_config
...
       def git_remotes
-        return [] if git_exe_path.nil?
-        `#{git_exe_path} remote --verbose`.to_s.strip.split("\n")
+        output = ""
+        begin
+          _process, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
+        rescue Errno::ENOENT => e
+          Jekyll.logger.warn "Not Installed:", e.message
+        end
+        output.to_s.strip.split("\n")
```
benbalter: Could this be simplified to...?

```ruby
def git_remotes
  _process, output = Jekyll::Utils::Exec.run("git", "remote", "--verbose")
  output.to_s.strip.split("\n")
rescue Errno::ENOENT => e
  Jekyll.logger.warn "Not Installed:", e.message
end
```

parkr: @ashmaroli Are you willing to update this PR to use @benbalter's suggestion?

spec/repository_finder_spec.rb
```diff
@@ -77,12 +77,14 @@
     end
 
     context "when git doesn't exist" do
-      before(:each) { @old_path = ENV.delete("PATH").to_s.split(File::PATH_SEPARATOR) }
-      after(:each)  { ENV["PATH"] = @old_path.join(File::PATH_SEPARATOR) }
+      before(:each) do
+        @old_path = ENV["PATH"]
+        ENV["PATH"] = ""
```
parkr: ?

spec/repository_finder_spec.rb
```diff
@@ -77,12 +77,14 @@
...
+      end
+      after(:each)  { ENV["PATH"] = @old_path }
 
       it "fails with a nice error message" do
         allow(subject).to receive(:git_remote_url).and_call_original
-        expect(subject.send(:git_exe_path)).to eql(nil)
```
parkr: I think we still need to be able to run git? If we're relying on Jekyll::Utils::Exec, then we'll definitely need to make sure that the exe can be found, right?

lib/jekyll-github-metadata/repository_finder.rb
parkr: > ~~@ashmaroli Are you willing to update this PR to use @benbalter's suggestion?~~

Oops! This is meant for @GICodeWarrior, not @ashmaroli  😄

spec/repository_finder_spec.rb
ashmaroli: Relying on `Jekyll::Utils::Exec` is like running `system "git"` in a `begin..rescue` block..
If the utility method doesn't find the executable, an `Errno::ENOENT` is raised which is rescued to output a warning to the terminal..

spec/repository_finder_spec.rb
ashmaroli: The `PATH` is "reset" to an empty string to simulate the scenario where `git` is not installed.. (especially in CI environments)

spec/repository_finder_spec.rb
parkr: Ok, so we can remove `git_exe_path`?

ashmaroli: Yep. In fact, this PR has already removed it..

parkr:
Thank you!

@jekyllbot: merge +bug

PR #139, opened 2018-10-22
Normalize whitespace across platforms

ashmaroli:
Attempt to fix EOL errors on AppVeyor

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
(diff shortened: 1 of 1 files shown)


pathawks:
Would it be better to do this in `.gitattributes`?

ashmaroli:
> Would it be better to do this in `.gitattributes`?

Perhaps. But I don't think Sublime Text and the like would pick that up.. not sure. Will have to check..

pathawks:
If you want to normalize `CRLF` vs `LF`, you can do that in `.gitattributes` and Git will take care of it all internally. If the goal has more to do with preserving whitespace at the end of lines for Markdown, this is the way to go :+1:

ashmaroli:
Guys, I'm merging this since `script/cibuild` will not run completely on AppVeyor with current `master` as relayed from existing AppVeyor logs..

PR #140, opened 2018-10-28
Remove redundant code

ashmaroli:
Some of the checks placed here were to support Jekyll 2.x and is no longer necessary since we're requiring `Jekyll 3.1+` now

Not sure why the following has a check since Jekyll has always had `Site` constant defined.. So leaving that as is..
https://github.com/jekyll/github-metadata/blob/2c350b2fea29ee5fe18243ff84040a755a9de508/lib/jekyll-github-metadata.rb#L37-L39

### DIFF
--- a/lib/jekyll-github-metadata.rb
+++ b/lib/jekyll-github-metadata.rb
@@ -44,11 +44,7 @@ module Jekyll
       end
 
       def logger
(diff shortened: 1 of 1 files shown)


parkr:
@jekyllbot: merge +minor

PR #151, opened 2019-01-03
Expose User/Org information under site.github.owner

### LINKED ISSUE #149: Please expose user information for User pages
I'm trying to access information about the GitHub user from a User gh-pages template, information such as the full name, location, company, etc.

More specifically, I would like to see the fields in [this response](https://developer.github.com/v3/users/#response).

I tried looking at `{% for user in site.github.organization_members %}`, expecting that would be the one user that owns the User pages, but that's empty... Then I tried `{% for user in site.github.contributors %}`, but that only lists the base information (such as the username), which might be OK if I had access to making more API requests, but from gh-pages I really can't...

Can we expose a field that would be only valid on User pages and that would be populated with that response for the owner of that site?

**NOTE:** I'm happy to try to contribute, though Ruby isn't really my thing and I'm unsure of how to actually test this... If you can give me some pointers, I can try to hack something together to make this work.

Thanks!
Filipe

filbranden:
Fixes #149.

Tested locally for a User object, captured everything I was interested in (full name, email, company, bio, ...)

/cc @parkr @benbalter 

Thanks!
Filipe

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -62,6 +62,10 @@ module Jekyll
         end
       end
 
+      def owner_metadata
+        memoize_value :@owner_metadata, Value.new(proc { |c| c.organization(owner) || c.user(owner) })
+      end
+
       def owner_url
         "#{Pages.github_url}/#{owner}"
       end
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -62,8 +62,62 @@ module Jekyll
         end
       end
 
+      # Whitelisted keys for Organizations and Users
+      WHITELISTED_ORGANIZATION_KEYS = Set.new([
+        :login,
+        :id,
+        :node_id,
+        :url,
+        :avatar_url,
+        :description,
+        :name,
+        :company,
+        :blog,
+        :location,
+        :email,
+        :is_verified,
+        :has_organization_projects,
+        :has_repository_projects,
+        :public_repos,
+        :public_gists,
+        :followers,
+        :following,
+        :html_url,
+        :created_at,
+        :type,
+        :collaborators,
+      ])
+
+      WHITELISTED_USER_KEYS = Set.new([
+        :login,
+        :id,
+        :node_id,
+        :avatar_url,
+        :html_url,
+        :type,
+        :site_admin,
+        :name,
+        :company,
+        :blog,
+        :location,
+        :bio,
+        :public_repos,
+        :public_gists,
+        :followers,
+        :following,
+        :created_at,
+        :updated_at,
+      ])
+
       def owner_metadata
-        memoize_value :@owner_metadata, Value.new(proc { |c| c.organization(owner) || c.user(owner) })
+        memoize_value :@owner_metadata, Value.new(proc { |c|
+          org = c.organization(owner)
+          if org
+            org.to_h.select { |k, _| WHITELISTED_ORGANIZATION_KEYS.include? k }
+          else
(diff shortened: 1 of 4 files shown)


filbranden:
Update: Fixed the test case, so now CI tests pass.

I also managed to test this on a project page (by setting env variable `PAGES_REPO_NWO`) and confirmed the organization object looks OK.

I investigated filtering fields such as email (using `c.user(owner, "type" => "public")`, similar to `c.list_repos()` below), but that does not work... If you really think that information should be filtered out, can you suggest a way to do that?

Thanks!
Filipe

parkr:
I think we'd feel safer with an allow list. This would also allow us to upgrade to GraphQL in the future. Here's what I think we should allow:

```json
{
  "login": "octocat",
  "id": 1,
  "node_id": "MDQ6VXNlcjE=",
  "avatar_url": "https://github.com/images/error/octocat_happy.gif",
  "html_url": "https://github.com/octocat",
  "type": "User",
  "site_admin": false,
  "name": "monalisa octocat",
  "company": "GitHub",
  "blog": "https://github.com/blog",
  "location": "San Francisco",
  "bio": "There once was...",
  "public_repos": 2,
  "public_gists": 1,
  "followers": 20,
  "following": 0,
  "created_at": "2008-01-14T04:33:35Z",
  "updated_at": "2008-01-14T04:33:35Z"
}
```

For organizations, let's allow only the following fields:

```json
{
  "login": "github",
  "id": 1,
  "node_id": "MDEyOk9yZ2FuaXphdGlvbjE=",
  "url": "https://api.github.com/orgs/github",
  "avatar_url": "https://github.com/images/error/octocat_happy.gif",
  "description": "A great organization",
  "name": "github",
  "company": "GitHub",
  "blog": "https://github.com/blog",
  "location": "San Francisco",
  "email": "octocat@github.com",
  "is_verified": true,
  "has_organization_projects": true,
  "has_repository_projects": true,
  "public_repos": 2,
  "public_gists": 1,
  "followers": 20,
  "following": 0,
  "html_url": "https://github.com/octocat",
  "created_at": "2008-01-14T04:33:35Z",
  "type": "Organization",
  "collaborators": 8,
}
```


What do you think?

filbranden:
Sure @parkr that list looks pretty reasonable. I can work on an update and push to this PR. Thanks!

filbranden:
Done. Please take a look.

The AppVeyor failures look unrelated... Can you please relaunch those two failing jobs?

ashmaroli:
> Can you please relaunch those two failing jobs?

@filbranden Done!

lib/jekyll-github-metadata/repository.rb
```diff
@@ -62,6 +62,64 @@ def repo_pages_info_opts
...
+        :bio,
+        :public_repos,
+        :public_gists,
+        :followers,
+        :following,
+        :created_at,
+        :updated_at,
+      ])
+
+      def owner_obj
```
parkr: Instead of owner_obj, how about calling this `owner_metadata`? A comment above the method would also help tremendously! I'm also tempted to move this user lookup & filter into a separate Drop class...

filbranden: Done.

(When first writing the code, I was thinking of a follow up that would rename `owner` to `owner_name` to make those match... It's a much more intrusive patch, but maybe it would be nice to have the same name on either side?)

In any case, renamed as requested. Please take another look.

filbranden:
AppVeyor is 1/4 unhappy... Please restart that build if possible? Thanks!

parkr:
Thank you so much!

@jekyllbot: merge +minor

PR #158, opened 2019-01-29
Add new attributes to return for users and repositories

kenyonj:
This PR:
- Adds `hireable` to the whitelisted attributes available on the `owner` hash.
- Adds decorator methods for `stargazers_count` and `forks_count` to the `Repository` object.

Why?
- This will allow users of this gem to access more useful information from users and repositories.

PR #159, opened 2019-01-29
Move owner_metadata to an Owner class and add specs

parkr:
This is a follow-on from https://github.com/jekyll/github-metadata/pull/158 regarding adding tests.

It also splits out the owner_metadata into an Owner class which can be fetched and attributes allowed as needed. ✨ 

/cc @kenyonj


lib/jekyll-github-metadata/owner.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      private
+
+      def owner_info
+        @owner_info ||= begin
+                          Value.new(proc { |c|
+                                      (c.organization(owner_login) || c.user(owner_login)).to_h
+                                    }).render || {}
+                        end
+      end
```
ashmaroli: Please format this method as:
```ruby
      def owner_info
        @owner_info ||= begin
          Value.new(proc { |c|
            (c.organization(owner_login) || c.user(owner_login)).to_h
          }).render || {}
        end
      end
```

parkr: @ashmaroli I can't! This is Rubocop doing it this way.

spec/owner_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      :is_verified               => true,
+      :has_organization_projects => true,
+      :has_repository_projects   => true,
+      :public_repos              => 50,
+      :public_gists              => 0,
+      :followers                 => 0,
+      :following                 => 0,
+      :html_url                  => "https://github.com/jekyll",
+      :created_at                => Time.parse("2012-12-19 19:37:35 UTC"),
+      :updated_at                => Time.parse("2019-01-27 15:27:32 UTC"),
```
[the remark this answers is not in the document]
parkr: Hm, good point. Let me see if I can come up with a better way to test this.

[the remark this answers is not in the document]
parkr: Ah! I see why I did it this way. These are all *fixtures*, so their values are fixed based on files in this repo. See `spec/webmock/*.json` 😄

parkr:
AppVeyor is failing because web mock appears not to be working?

ashmaroli:
I've restarted the failed jobs. Should pass now..

lib/jekyll-github-metadata/owner.rb
```diff
@@ -0,0 +1,86 @@
...
+                          :location,
+                          :login,
+                          :name,
+                          :node_id,
+                          :public_gists,
+                          :public_gists,
+                          :public_repos,
+                          :public_repos,
+                          :type,
+                          :updated_at
```
ashmaroli: @parkr  if you don't mind, *may I collapse this list..* to multiple attributes per line?

lib/jekyll-github-metadata/owner.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+        end
+      end
+
+      def self.content_methods
+        @content_methods ||= []
+      end
+
+      # List of whitelisted keys.
+      def_hash_delegators :owner_info,
```
ashmaroli: Since there is also a `private` method with the same name as `:owner_info`, which definition takes precedence?

lib/jekyll-github-metadata/owner.rb
```diff
@@ -0,0 +1,86 @@
...
+      # Defines an instance method that delegates to a hash's key
+      #
+      # hash   - a symbol representing the instance method to delegate to. The
+      #          instance method should return a hash or respond to #[]
+      # key    - the key to call within the hash
+      # method - (optional) the instance method the key should be aliased to.
+      #          If not specified, defaults to the hash key
+      #
+      # Returns a symbol representing the instance method
+      def self.def_hash_delegators(hash, *methods)
```
benbalter: Since I introduced this, glad to move this to a module in a subsequent PR to DRY things up.

lib/jekyll-github-metadata/owner.rb
parkr: @ashmaroli I'd prefer each attribute to be on their own lines. The reason is that diffs are *much* clearer when we're working with whole lines. This is also an allow list, so for auditing purposes, it makes it a little easier to look at. 😄

lib/jekyll-github-metadata/owner.rb
parkr: Not a huge deal to me. This is a little different than the Repository class. 🤷‍♂️

parkr:
@jekyllbot: merge +minor

PR #163, opened 2019-03-17
Conditionally memoize certain private methods in EditLinkTag

ashmaroli:
Closes #162 

## Summary

Jekyll 4.0 caches parsed Liquid Templates. Therefore, when `{% github_edit_link %}` is used through a layout or an include, there is just *one instance of `Jekyll::GitHubMetadata::EditLinkTag`* in memory.

This poses a problem when attributes required for rendering the tag are memoized, resulting in the output of the *same result* for every page.

The changes proposed here allows compatibility with Jekyll 3.x where memoization improves performance and avoids memoizing when using Jekyll 4.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ matrix:
   include:
     - rvm: *latest_ruby
       env: JEKYLL_VERSION=3.7.0
+    - rvm: *latest_ruby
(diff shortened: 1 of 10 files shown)


ashmaroli:
@jekyllbot: merge +dev

PR #170, opened 2019-09-20
Support DotEnv in all non-production environments (fixes #168)

### LINKED ISSUE #168: Allow DotEnv in environment other than development
Currently, DotEnv is only loaded when `Jekyll.env == 'development'`. However, thanks to jekyll/jekyll#5431, `site.url` config is set to `http://host:port` in development environment, rather than respecting site configuration. This means that it's not possible to use a URL from config file in conjunction with `.env` file.

Since environment can be any value, suggestion would be to change the check to `Jekyll.env != 'production'`. This would allow something like the following:

```JEKYLL_ENV=test bundle exec jekyll serve```

Here, `.env` file would be loaded, but `site.url` would still be respected.

cmbuckley:
See #168


asbjornu:
@mattr- Any reason this isn't merged?

PR #173, opened 2019-10-02
Fix test failures

parkr:
* MetadataDrop: update mutations reference to use helper method
* Lock rspec-expectations and rspec-mocks


ashmaroli:
@parkr Something seems to have changed with either the `rspec-expectations` gem or perhaps gem `rspec-mocks`, both of which saw a release couple of hours ago today.

I've pushed a commit to lock the versions of these gems from the last known passing build.

PR #189, opened 2020-07-02
Add more log statements to site_github_munger.

NullableInt:
This adds more logs that were critical for us to solve an deploy error.

### DIFF
--- a/lib/jekyll-github-metadata/site_github_munger.rb
+++ b/lib/jekyll-github-metadata/site_github_munger.rb
@@ -42,10 +42,14 @@ module Jekyll
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
+        Jekyll::GitHubMetadata.log :debug, "Adding url and base url fallbacks..."
         site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })
+          
+        Jekyll::GitHubMetadata.log :debug, "Url is set to #{site.config["url"]}"
         return unless should_set_baseurl?
 
         site.config["baseurl"] = Value.new("baseurl", proc { |_c, r| r.baseurl })
+        Jekyll::GitHubMetadata.log :debug, "baseurl is set to #{site.config["baseurl"]}"
       end
 
       def add_title_and_description_fallbacks!


lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -42,10 +42,14 @@ def drop
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
+        Jekyll::GitHubMetadata.log :debug, "Adding url and base url fallbacks..."
         site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })
+          
+        Jekyll::GitHubMetadata.log :debug, "Url is set to #{site.config["url"]}"
```
MichaelCurrin, line 48:
```suggestion
        Jekyll::GitHubMetadata.log :debug, "URL is set to #{site.config["url"]}"
```

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -42,10 +42,14 @@ def drop
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
+        Jekyll::GitHubMetadata.log :debug, "Adding url and base url fallbacks..."
```
MichaelCurrin, line 45: This log applies not just for the line immediately after it but the entire function `add_url_and_baseurl_fallbacks`, so it is clearer to have an empty line after it.

```suggestion
        Jekyll::GitHubMetadata.log :debug, "Adding URL and base URL fallbacks..."

```

Also maybe take this whole log line out? I don't know if this line actually tells you anything usefu,l as it doesn't tell you actual values. And if there was an error in one of the lines in the function, that would be clear without having to see a log that told you that you were entering function.

lib/jekyll-github-metadata/site_github_munger.rb
MichaelCurrin: Also this line has more to do with `site.config["url"] ||...` than `return unless should_set_baseurl?` so why not swap this line with the blank line above it?

MichaelCurrin (changes requested):
Update log messages

lib/jekyll-github-metadata/site_github_munger.rb
MichaelCurrin: The original issue has this as output:

```
GitHub Metadata: Adding url and base url fallbacks...
GitHub Metadata: Url is set to https://github.com
GitHub Metadata: baseurl is set to /pages/SwedbankPay/swedbank-pay-design-guide-jekyll-theme
```

But this would be acceptable and useful I think:

```
GitHub Metadata: Url is set to https://github.com
GitHub Metadata: baseurl is set to /pages/SwedbankPay/swedbank-pay-design-guide-jekyll-theme
```

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -42,10 +42,13 @@ def drop
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
-        site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })
+        Jekyll::GitHubMetadata.log :debug, "Adding URL and base URL fallbacks..."
```
MichaelCurrin, line 45: This describes the whole function, so some whitespace is useful, so indicate it applies to the section and not just the line or two directly after it.
```suggestion
        Jekyll::GitHubMetadata.log :debug, "Adding URL and base URL fallbacks..."

```

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -42,10 +42,13 @@ def drop
...
+        site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })          
+        Jekyll::GitHubMetadata.log :debug, "URL is set to #{site.config["url"]}"
         return unless should_set_baseurl?
 
         site.config["baseurl"] = Value.new("baseurl", proc { |_c, r| r.baseurl })
+        Jekyll::GitHubMetadata.log :debug, "baseurl is set to #{site.config["baseurl"]}"
```
MichaelCurrin, line 51: Use `base URL` here for consistency with the above.

---

Or use `url` and `baseurl` (lowercase, no spaces) throughout.

Some quotes or backticks would be slightly better then.

```
"`baseurl` is set to #{site.config["baseurl"]}"
```

```
"'baseurl' is set to #{site.config["baseurl"]}"
```

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -42,10 +42,13 @@ def drop
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
-        site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })
+        Jekyll::GitHubMetadata.log :debug, "Adding url and baseurl fallbacks..."
```
MichaelCurrin, line 45:
```suggestion
        Jekyll::GitHubMetadata.log :debug, "Adding `url` and `baseurl` fallbacks..."
        
```

MichaelCurrin (changes requested):
Apply suggestions consistently

MichaelCurrin (approved):
Thanks, changes look good

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -42,10 +42,14 @@ def drop
 
       # Set `site.url` and `site.baseurl` if unset.
       def add_url_and_baseurl_fallbacks!
-        site.config["url"] ||= Value.new("url", proc { |_c, r| r.url_without_path })
+        Jekyll::GitHubMetadata.log :debug, "Adding `url` and `baseurl` fallbacks..."
```
parkr, line 45: These would be better in the `Value` `proc { }` block, otherwise the benefits of using a Value (the fact that computation is deferred) is lost.

PR #191, opened 2020-10-29
set PAGES_GITHUB_HOSTNAME to hostname without protocol

Shegox:
This PR sets the `PAGES_GITHUB_HOSTNAME` example to `github.com` without the protocol as GitHub pages also sets the `site.github.hostname` without the protocol.


parkr:
Thank you!

@jekyllbot: merge +docs

PR #197, opened 2021-02-23
Use owner name as site title for User and Organization sites.

### LINKED ISSUE #196: Use owner name as site title for User and Organization sites.
Using the repo name as the site title makes sense for Projects, which can choose their own repo names. But User and Organization sites must use repo names like `{owner}.github.io` which:

* Does not look good as a browser tab title, bookmark, or header for a page that is supposed to be about a person or organization.
* Conveys no useful information, as it is unlikely that anyone would clone or fork such a repo.
* Looks confusing for users or organizations that use custom domains (CNAMEs), and have no other content indicating their site is hosted by GitHub.

The display name of the user or organization is a better choice. This is an optional property, so fall back to (login) name if display name is not present.

lardieri:
Resolve #196

Using the repo name as the site title makes sense for Projects, which can choose their own repo names. But User and Organization sites must use repo names like `{owner}.github.io` which:

* Does not look good as a browser tab title, bookmark, or header for a page that is supposed to be about a person or organization.
* Conveys no useful information, as it is unlikely that anyone would clone or fork such a repo.
* Looks confusing for users or organizations that use custom domains (CNAMEs), and have no other content indicating their site is hosted by GitHub.

The display name of the user or organization is a better choice. This is an optional property, so fall back to (login) name if display name is not present.


lardieri:
I'd like to add a couple of new tests in site_github_munger_spec.rb around line 110 or so, but it looks like I might need to add a new test-site as well. Probably best to consult with a project owner before proceeding.

mattr-:
Add your tests and a new test site. Whatever you think is necessary to properly test this.

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -55,7 +55,13 @@ def add_title_and_description_fallbacks!
...
           Jekyll::GitHubMetadata.log :warn, msg
         else
-          site.config["title"] ||= Value.new("title", proc { |_c, r| r.name })
+          site.config["title"] ||= Value.new("title", proc { |_c, r|
+            if r.project_page?
+              r.name # Someday if GitHub supports display names for repos, prefer it here.
+            else
+              r.owner_displayname || r.owner
+            end
+          })
```
mattr-, line 64: What do you think about expanding the variable names here? I understand that you're just rolling with what was already here, but I think using `_context` and `repository` instead of `_c` and `r` respectively would be a nice readable improvement.

mattr- (approved):
I think this is a nice addition as is, but i support your desire to add tests. Please add whatever tests and related fixtures (test site, etc.) that you think are necessary. However, if you find after a few days that adding the tests is a lot of extra work, let's go ahead and skip them and ship this without.

lib/jekyll-github-metadata/site_github_munger.rb
lardieri: Sure, that sounds good. Should I change the entire file, for consistency? If so, I might want to submit a separate PR so as to not noise-up the diff.

mattr-: Since you're changing just this code here already, let's only do this part.

lib/jekyll-github-metadata/metadata_drop.rb
```diff
@@ -48,6 +48,7 @@ def keys
       def_delegator :repository, :tagline,                     :project_tagline
       def_delegator :repository, :owner_metadata,              :owner
       def_delegator :repository, :owner,                       :owner_name
+      def_delegator :repository, :owner_displayname,           :owner_displayname
```
MichaelCurrin, line 51: How about `owner_display_name` throughout?

Username and metadata do make sense as single words.

But it would be "display name" as two words not "displayname", outside of the code. Therefore underscore in the code would split the words well.

MichaelCurrin (changes requested):

MichaelCurrin:
CI checks are failing still.

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -55,7 +55,13 @@ def add_title_and_description_fallbacks!
           msg << "Jekyll GitHub Metadata will not set site.title to the repository's name."
           Jekyll::GitHubMetadata.log :warn, msg
         else
-          site.config["title"] ||= Value.new("title", proc { |_c, r| r.name })
+          site.config["title"] ||= Value.new("title", proc { |_context, repository|
+            if repository.project_page?
+              repository.name # Someday if GitHub supports display names for repos, prefer it here.
```
MichaelCurrin, line 60: Rather remove this TODO.

It's clear from the implementation that only `name` is being used not a display name like below.

If display name ever became a thing, there's be an issue to investigate areas to use it and this area of `repository.name` should come up easily without the comment adding clutter.

.gitignore
```diff
@@ -5,3 +5,4 @@ pkg
 _site
 tmp
 *.gem
+.jekyll-cache
```
MichaelCurrin, line 8: This could have been better handled in a separate issue/PR and discussed separately.
But this fine to include as it supports Jekyll 4 cache.

MichaelCurrin:
Thanks for renaming.

Please see additional suggestion on removing the code comment.

Also please fix the 2 failing AppVeyor tests from _checks_.

Well, the one looks like SSL error outside of this PR scope.

The other issue looks more relevant. [Test job](https://ci.appveyor.com/project/jekyll/github-metadata/builds/37997173/job/658v5ai205upmw7j) from [build #37997173](https://ci.appveyor.com/project/jekyll/github-metadata/builds/37997173)


          1) Jekyll::GitHubMetadata::Owner is an ORG fetches login
             Failure/Error: expect(subject.public_send(attribute)).to eq(expected_value)
               expected: "jekyll"
                    got: nil
               (compared using ==)
             # ./spec/owner_spec.rb:43:in `block (4 levels) in <top (required)>'
        Finished in 1 minute 31.39 seconds (files took 1.23 seconds to load)

MichaelCurrin:
## Tests

I see two 2 failing AppVeyor tests from _checks_.

Tthe one looks like SSL error outside of this PR scope.

The other issue looks possibly relevant. [Test job](https://ci.appveyor.com/project/jekyll/github-metadata/builds/37997173/job/658v5ai205upmw7j) from [build #37997173](https://ci.appveyor.com/project/jekyll/github-metadata/builds/37997173)


          1) Jekyll::GitHubMetadata::Owner is an ORG fetches login
             Failure/Error: expect(subject.public_send(attribute)).to eq(expected_value)
               expected: "jekyll"
                    got: nil
               (compared using ==)
             # ./spec/owner_spec.rb:43:in `block (4 levels) in <top (required)>'
        Finished in 1 minute 31.39 seconds (files took 1.23 seconds to load)

But also, tests are failing in general across PRs so this may have been an issue already.

https://ci.appveyor.com/project/jekyll/github-metadata/history

MichaelCurrin (changes requested):
Remove comment

lardieri:
> ## Tests
> The other issue looks possibly relevant. [Test job](https://ci.appveyor.com/project/jekyll/github-metadata/builds/37997173/job/658v5ai205upmw7j) from [build #37997173](https://ci.appveyor.com/project/jekyll/github-metadata/builds/37997173)

My new code does not distinguish between User and Organization owners. If my code caused this failure, then the test case `Jekyll::GitHubMetadata::Owner is a USER` would have failed as well.

> But also, tests are failing in general across PRs so this may have been an issue already.
> 
> https://ci.appveyor.com/project/jekyll/github-metadata/history

The failures are not consistent across obviously irrelevant commits, such as my change to `.gitignore`.

Nor are the failures reproducible locally using the same seed and the same Ruby / Jekyll versions in the environment.

It's reasonable to conclude that some issue with the AppVeyor environment is responsible for the failures. Possible hypotheses:

- AppVeyor is not pulling the environment that it says that it is.
-- Perhaps a corrupt local cache of some kind.
-- Perhaps some bug or other difference in the Windows implementation of Ruby vs. Unix.
- Hardware issue with AppVeyor worker machines (random 1-bit errors, etc.).
- Some type of Chaos Monkey has been introduced.

There are also obvious issues with order dependency in these tests. For example:

- `--seed 41261` produces warnings for test cases `when git doesn't exist` and `with name set, but no title`
- `--seed 58031` runs the same tests in a different order and produces no warnings

(I would argue that the warnings are correct, given the context of those specific cases.)

So, I'm not losing sleep over these test failures. If you disagree, please suggest a concrete course of investigation.

MichaelCurrin:
Thanks for investigating. Yes I suspected the tests were in an inconsistent state already so further changes to tests aren't need I think

lardieri:
Might also resolve #207 ?

lardieri:
@mattr-  I just pushed the test cases that I promised you back in February. Would you mind taking a look, and merging if you approve? Thank you.

lardieri:
@mattr- and @MichaelCurrin  - any further thoughts on this? or are you ready to merge it in? Thanks.

lib/jekyll-github-metadata/repository.rb
```diff
@@ -64,6 +64,10 @@ def repo_pages_info_opts
         end
       end
 
+      def owner_display_name
+        owner_metadata.name
```
MichaelCurrin, line 68:
```suggestion
        @owner_metadata.name
```

Should this be `@` for consistency with the code below it? Unless it is different for a reason?

```
def owner_metadata
        @owner_metadata ...
```

MichaelCurrin: Oh nevermind. I see it uses the attribute wrapper `owner_metadata`, so we set `@owner_metadata` if not set already.

MichaelCurrin (approved):

MichaelCurrin:
@lardieri The code and tests look clear and thorough. Approved.

mattr-:
@jekyllbot: merge +minor

PR #202, opened 2021-03-08
Restore log level after running tests that modify it.

### LINKED ISSUE #201: Restore log level after running tests that modify it.
Certain tests in the suite deliberately provoke warnings.

* site_github_munger_spec.rb:
`GitHub Metadata: site.name is set in _config.yml, but many plugins and themes expect site.title to be used instead.`
* repository_finder_spec.rb:
`Not Installed: No such file or directory - git`

However, other tests set the log level to Error, which suppresses warnings:
* edit_link_tag_spec.rb
* integration_spec.rb

Rspec randomizes the order of the tests. If a random seed is chosen that orders the warning tests ahead of the error tests, then the warnings get logged correctly. Example: `./scripts/test --seed 41261`

However, if the random seed orders either of the error tests ahead of the warning tests, then the warnings get suppressed. Example: `./scripts/test --seed 60792`

Tests that modify the log level should restore it upon completion.

lardieri:
Resolve #201

Certain tests in the suite deliberately provoke warnings.

* site_github_munger_spec.rb:
`GitHub Metadata: site.name is set in _config.yml, but many plugins and themes expect site.title to be used instead.`
* repository_finder_spec.rb:
`Not Installed: No such file or directory - git`

However, other tests set the log level to Error, which suppresses warnings:
* edit_link_tag_spec.rb
* integration_spec.rb

Rspec randomizes the order of the tests. If a random seed is chosen that orders the warning tests ahead of the error tests, then the warnings get logged correctly. Example: --seed 41261

However, if the random seed orders either of the error tests ahead of the warning tests, then the warnings get suppressed. Example: --seed 60792

This commit makes the error tests remember the original log level and restore it after they run. Now the warnings will get logged every time.

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -14,12 +14,14 @@ RSpec.describe("integration into a jekyll site") do
     ENV.delete("JEKYLL_ENV")
     ENV["JEKYLL_ENV"] = "production"
     ENV["PAGES_ENV"] = "dotcom"
+    @original_log_level = Jekyll.logger.level
     Jekyll.logger.log_level = :error
     Jekyll::Commands::Build.process(config_defaults)
   end
   after(:each) do
     ENV.delete("PAGES_ENV")
     ENV["JEKYLL_ENV"] = "test"
+    Jekyll.logger.log_level = @original_log_level
   end
   subject { SafeYAML.load(in_dest_dir("rendered.txt").read) }
 
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -14,14 +14,12 @@ RSpec.describe("integration into a jekyll site") do
     ENV.delete("JEKYLL_ENV")
     ENV["JEKYLL_ENV"] = "production"
     ENV["PAGES_ENV"] = "dotcom"
-    @original_log_level = Jekyll.logger.level
     Jekyll.logger.log_level = :error
     Jekyll::Commands::Build.process(config_defaults)
   end
   after(:each) do
     ENV.delete("PAGES_ENV")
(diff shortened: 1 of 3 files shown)


lardieri:
All three expected test warnings are present in the console log for the most recent AppVeyor run above:
[https://ci.appveyor.com/project/jekyll/github-metadata/builds/38109706/job/k077phiqhin9mn82]

If you run the test suite on the master branch with the same seed (`--seed 47529`), the warnings are suppressed. So you can see the effectiveness of this change.

lardieri:
@mattr- @parkr How do you feel about this change?

spec/integration_spec.rb
```diff
@@ -14,12 +14,14 @@
...
     ENV["JEKYLL_ENV"] = "production"
     ENV["PAGES_ENV"] = "dotcom"
+    @original_log_level = Jekyll.logger.level
     Jekyll.logger.log_level = :error
     Jekyll::Commands::Build.process(config_defaults)
   end
   after(:each) do
     ENV.delete("PAGES_ENV")
     ENV["JEKYLL_ENV"] = "test"
+    Jekyll.logger.log_level = @original_log_level
```
parkr, line 24: What do you think about moving this after and the before as well into the spec_helper.rb file? That way every spec can have the log level reset, so all they need to do is set a new log level as desired.

lardieri: @parkr I agree, that is a good idea. I'm pushing a commit to do this. Please let me know what you think.

parkr (approved):

parkr:
Thank you so much!

@jekyllbot: merge +dev

PR #219, opened 2021-12-22
Get path for first page of paginated pages (fixes #213)

### LINKED ISSUE #213: github_edit_link contains pagination path
If a page is `paginated: true`, all pages (except the first) contain the `paginate_path` in the `github_edit_link`.

I created a [minimal demo here](https://github.com/cmbuckley/github-edit-link-issue), but if you have the following config:

```
repository: "owner/repo"
paginate_path: "page/:num/"
```

and an index.html with `paginated: true`, then when you visit /page/2/ the `github_edit_link` will contain **/page/2/index.html** instead of just **/index.html**.

The issue here would seem to come from [jekyll-paginate](https://github.com/jekyll/jekyll-paginate), because the `page.path` is set to **/page/2/index.html**. It seemed better to raise the issue here since that repo is not maintained.

cmbuckley:
When a page is paginated, the edit link includes the pagination part of the URL. This grabs the path from the first page of the pagination collection.

I would appreciate some help with the tests as I'm not familiar enough with Ruby.

### DIFF
--- a/lib/jekyll-github-metadata/edit-link-tag.rb
+++ b/lib/jekyll-github-metadata/edit-link-tag.rb
@@ -35,7 +35,6 @@ module Jekyll
       private def_hash_delegator :site_github, :source,         :source, {}
       private def_hash_delegator :source,      :branch,         :branch
       private def_hash_delegator :source,      :path,           :source_path
-      private def_hash_delegator :page,        :path,           :page_path
 
       def render(context)
         @context = context
@@ -70,6 +69,16 @@ module Jekyll
 
       def parts
         memoize_conditionally { [repository_url, "edit/", branch, source_path, page_path] }
+      end
+
+      def page_path
+        return page['path'] unless page['paginated']
+
+        site.pages.dup.select do |page|
+          if page.pager and page.pager.page == 1
+            return page['path']
+          end
+        end
       end
 
       def parts_normalized


lib/jekyll-github-metadata/edit-link-tag.rb
```diff
@@ -72,6 +71,14 @@ def parts
         memoize_conditionally { [repository_url, "edit/", branch, source_path, page_path] }
       end
 
+      def page_path
+        return page["path"] unless page["paginated"]
+
+        site.pages.dup.select do |page|
```
ashmaroli, line 77: There's no need to use `.dup` here.
`Array#select` does not mutate the original array.

Additionally, since you're not using the *filtered* result of `Array#select`, you might as well iterate directly
```suggestion
        site.pages.each do |page|
```

cmbuckley: Fair enough, I’d just lifted that from jekyll-paginate so wasn't sure, but I see [in an unreleased commit they had done the same](https://github.com/jekyll/jekyll-paginate/commit/70efd9695cb49157fbc74509d5eb2e48ee266252). I had figured `.dup` would reduce the number of pages iterated over to find the right path?

ashmaroli: > I had figured `.dup` would reduce the number of pages iterated over to find the right path?

Nope. `.dup` simply allocates a *duplicate* copy of the original array. Another *duplicate* gets created by the subsequent `.select`.

cmbuckley: Thanks for the explanation — having only seen `.clone` in the past, I had mistaken `.dup` for something intrinsic to Jekyll.

ashmaroli:
Okay. The existing tests pass.  Therefore, no regressions.
Now you need to add the test that proves this change works (with your use-case) as claimed.

cmbuckley:
Hmm, not what I was expecting. Although I note that the tests are running against Jekyll 4.2.1 and I was using 3.9.0 to match the version in `github-pages` gem.

spec/spec_helpers/fixture_helper.rb
```diff
@@ -22,8 +22,15 @@ def config_defaults
     }
   end
 
-  def make_page(data = {})
-    Jekyll::Page.new(site, config_defaults["source"], "", "page.md").tap { |page| page.data = data }
+  def make_page(data = {}, pager = nil)
+    data["paginated"] = true unless pager == nil
+    page = Jekyll::Page.new(site, config_defaults["source"], "", "page.md").tap { |page| page.data = data }
```
ashmaroli, line 27: FYI, `.tap` is used in lieu of assigning a local variable.
Using both is just smelly code.

parkr:
Seems like a useful bug fix.

Issue #233, opened 2022-04-29
any plan to implement site.github.public_repositories[].contributors ?


9bow:
Hello,

I'm running a GitHub organization, and I would like to publish contributor list of all public repos in organization to our website.

In this regard, I was expecting the following code, but the field `site.github.public_repositories[].contributors` was not exist.

```
{% for repository in site.github.public_repositories %}
  ## [{{ repository.name }}]({{ repository.html_url }})
  {% for contributor in repository.contributors %}
    * [{{ contributor.name }}]({{ contributor.html_url }})
  {% endfor %}
{% endfor %}
```

I found [an old similar issue](https://github.com/jekyll/github-metadata/issues/26) with this feature, but it seems to have been resolved by fetching only the contributors of only one repository (the current repository specified in `_config.yaml`) as `site.github.contributors`.

It looks like the [GitHub API](https://docs.github.com/en/rest/repos/repos#list-repository-contributors) and [octokit.rb](https://github.com/octokit/octokit.rb/blob/d1628ebd3e71aeb1534b3115c8bf37e8fb42068f/lib/octokit/client/repositories.rb#L469) support the feature to get a list of contributors by repository.

However, I'm a bit embarrassed because I haven't written code in ruby. :sweat_smile:

Do you have any plans to add a feature related to this?

Thank you.

9bow:
New feature added as PR (#234) to close this issue.
This feature has been merged with the dedicated help of maintainer @parkr.
With his review, this PR was a great experience for me. :)

parkr:
Thank you for the contribution, @9bow!

PR #224, opened 2022-04-15
Add site.github.public_repositories[].releases

parkr:
For each public repository rendered, we will also list releases for them.

Fixes #221.

PR #234, opened 2022-04-29
Add site.github.public_repositories[].contributors

9bow:
Hello, 

After I wrote #233, I saw a merged PR about `site.github.public_repositories[].releases` (#224 by @parkr )

I've never used ruby before, but with help of my friend @peniar , I made this PR. (thanks, @peniar)

Here're what I did:

First, I changed the code `lib/jekyll-github-metadata/repository.rb` in my installed gem, and check it worked.

And then, I wrote some spec code referring previous PR about release, assumed that return should be like `spec/webmock/api_get_repo_contributors.json`

Lastly, I tested with `bundle exec rspec spec`, and then  1 test I added failed as follows:
```
Failures:

  1) integration into a jekyll site calls all the stubs
     Failure/Error: expect(stub).to have_been_requested

       The request GET https://api.github.com/repos/jekyll/github-metadata/contributors?per_page=100 with headers {'Accept'=>/application\/vnd\.github\.(v3|drax-preview|mercy-preview)\+json/, 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'token 1234abc', 'Content-Type'=>'application/json', 'User-Agent'=>'Octokit Ruby Gem 4.22.0'} was expected to execute 1 time but it executed 2 times
```

I think the error `expected to execute 1 time but it executed 2 times` is because the line 16 in stub_helper.rb conflict with line 26 which I added. However, line 15 for releases code does not conflict with line 25. :(

I'm stuck here and don't know what to try or look for anymore.

If you could point out this PR, I would be very grateful.

### DIFF
--- a/lib/jekyll-github-metadata/repository.rb
+++ b/lib/jekyll-github-metadata/repository.rb
@@ -119,6 +119,7 @@ module Jekyll
         }
         memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc do |c|
           c.list_repos(owner, options).each { |r| r[:releases] = c.releases(r[:full_name]) }
+          c.list_repos(owner, options).each { |r| r[:contributors] = c.contributors(r[:full_name], options) }
         end)
       end
 
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -42,6 +42,14 @@ RSpec.describe("integration into a jekyll site") do
     expect(subject["public_repositories"].first["releases"].first["target_commitish"]).to eql("master")
   end
 
+  it "contains the correct public_repositories.contributors" do
+    expect(subject).to have_key("public_repositories")
+    expect(subject["public_repositories"].first).to have_key("contributors")
+    expect(subject["public_repositories"].first["contributors"].size).to eql(1)
+    expect(subject["public_repositories"].first["contributors"].first["login"]).to eql("parkr")
+    expect(subject["public_repositories"].first["contributors"].first["id"]).to eql(237985)
+  end
+
   it "calls all the stubs" do
     stubs.each do |stub|
       expect(stub).to have_been_requested
--- a/spec/spec_helpers/stub_helper.rb
+++ b/spec/spec_helpers/stub_helper.rb
@@ -23,6 +23,7 @@ module StubHelper
 
     owner_repos = JSON.parse(webmock_data("owner_repos"))
     owner_repos.each { |r| stubs << stub_api("/repos/#{r["full_name"]}/releases?per_page=100", "repo_releases") }
+    owner_repos.each { |r| stubs << stub_api("/repos/#{r["full_name"]}/contributors?per_page=100", "repo_contributors") }
 
     stubs
   end


lib/jekyll-github-metadata/repository.rb
```diff
@@ -119,6 +119,7 @@ def owner_public_repositories
         }
         memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc do |c|
           c.list_repos(owner, options).each { |r| r[:releases] = c.releases(r[:full_name]) }
+          c.list_repos(owner, options).each { |r| r[:contributors] = c.contributors(r[:full_name], options) }
```
parkr, line 122: This will make a second request to list repos, so I think you just want to do this to prevent that extra API request:

```suggestion
          c.list_repos(owner, options).each do |r| 
            r[:releases] = c.releases(r[:full_name])
            r[:contributors] = c.contributors(r[:full_name], options)
          end
```

parkr: I wonder if we can make these `Value`s so they only resolve (make the API request) if they're called. Otherwise we have these 2 (or many more, since auto_paginate is on) extra requests for every repo if any public repos are used in the site.

parkr (commented):
Definite 👍 to the idea... wondering if we can use the Value class to improve the performance / reduce API calls if we aren't going to use these fields.

lib/jekyll-github-metadata/repository.rb
9bow: Thanks! Actually I tried this style in first phase as below, but the code I tried doesn't return any key releases or contributors! Thankfully, I think I know a bit more ruby syntax.😭

```ruby
        memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc do |c|
          # c.list_repos(owner, options).each { proc do |r|
          #   r[:releases] = c.releases(r[:full_name])
          #   r[:contributors] = c.contributors(r[:full_name], options)
          #   r
          # end }
        end)
```

9bow: I totally agree. Actually in my case I don't need `release`, but others have tons of repo and contributors and I don't think they want this feature. :sweat_smile:
I will a little more research on what values can be passed in jekyll's `_config.yaml`.

9bow:
Thanks for your review, @parkr 

How about the following method? I made it work only when `JEKYLL_GITHUB_METADATA_DETAILS` is set to `true` by using an environment variable.
If you have any ideas on how to pass this flag or name convention, it would be nice if you could let me know. :)

```ruby
      def owner_public_repositories
        options = {
          :type   => "public",
          :accept => "application/vnd.github.mercy-preview+json",
        }
        if ENV["JEKYLL_GITHUB_METADATA_DETAILS"] == true
          memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc do |c|
            c.list_repos(owner, options).each do |r|
              r[:releases] = c.releases(r[:full_name])
              r[:contributors] = c.contributors(r[:full_name], options)
            end
          end)
        else
          memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc { |c| c.list_repos(owner, options) })
        end
      end
```

parkr:
I was able to get this to pass on my machine with the following change:

```diff
diff --git lib/jekyll-github-metadata/repository.rb lib/jekyll-github-metadata/repository.rb
index 1e4864c..ba49d95 100644
--- lib/jekyll-github-metadata/repository.rb
+++ lib/jekyll-github-metadata/repository.rb
@@ -118,9 +118,11 @@ module Jekyll
           :accept => "application/vnd.github.mercy-preview+json",
         }
         memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc do |c|
-          c.list_repos(owner, options).each do |r| 
-            r[:releases] = c.releases(r[:full_name])
-            r[:contributors] = c.contributors(r[:full_name], options)
+          c.list_repos(owner, options).map do |repo|
+            r = Sanitizer.sanitize(repo)
+            r[:releases] = Value.new("owner_public_repositories_releases", proc { c.releases(r["full_name"]) })
+            r[:contributors] = Value.new("owner_public_repositories_contributors", proc { c.contributors(r["full_name"]) })
+            r
           end
         end)
       end
diff --git lib/jekyll-github-metadata/sanitizer.rb lib/jekyll-github-metadata/sanitizer.rb
index edaf554..0a993d3 100644
--- lib/jekyll-github-metadata/sanitizer.rb
+++ lib/jekyll-github-metadata/sanitizer.rb
@@ -16,7 +16,7 @@ module Jekyll
       # resource - an Object
       #
       # Returns the sanitized resource.
-      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
       def sanitize(resource)
         case resource
         when Array
@@ -31,6 +31,8 @@ module Jekyll
           nil
         when String
           resource
+        when Value
+          resource.render
         else
           if resource.respond_to?(:to_hash)
             sanitize_resource(resource)
@@ -39,7 +41,7 @@ module Jekyll
           end
         end
       end
-      # rubocop:enable Metrics/CyclomaticComplexity
+      # rubocop:enable Metrics/CyclomaticComplexity, Metrics/MethodLength
 
       # Sanitize the Sawyer Resource or Hash
       # Note: the object must respond to :to_hash for this to work.
```

I don't think we want to make this an optional change – we can just load only when requested using `Value`.

9bow:
Thanks for the detailed guide, @parkr!

As per your guide, I wrapped the existing codes to use `Value.new()` and changed the `Sanitizer` module as well. I definetly agree that we do not want to make this an optional change, but honestly I'm still not sure what `Value` does. 😅 I think I need to study the whole code as well as the syntax of ruby more.

Anyway, please let me know if there is anything else I need to change or do.

9bow:
(TMI: My friend @peniar received e-mail when I commented, and he also taught me how `Value` works and how/why to use it. Thanks, @peniar!)

spec/integration_spec.rb
```diff
@@ -42,6 +42,14 @@
     expect(subject["public_repositories"].first["releases"].first["target_commitish"]).to eql("master")
   end
 
+  it "contains the correct public_repositories.contributors" do
+    expect(subject).to have_key("public_repositories")
+    expect(subject["public_repositories"].first).to have_key("contributors")
+    expect(subject["public_repositories"].first["contributors"].size).to eql(1)
+    expect(subject["public_repositories"].first["contributors"].first["login"]).to eql("parkr")
+    expect(subject["public_repositories"].first["contributors"].first["id"]).to eql(237985)
```
parkr, line 50: Rubocop is failing on this line. I think it's unnecessary so I'd recommend that you remove this line.

parkr (commented):
Two quick rubocop failures. Thank you!

9bow:
I just removed spec related codes, and am waiting for CI results!

parkr:
@9bow Sorry, I just meant the one line, not all your specs! I like your tests. Just remove the one line I pointed out, and combine the loop you added in spec_helper with the one right above it.

9bow:
Oh, I'm sorry @parkr. I misunderstood.
I'll revive those Spec codes and re-apply your review.

9bow:
I've been re-applied your review. Also verified that it passed the test locally as follows:
image

Sorry for the hassle and thank you for your kind and patient review!

parkr (approved):
Thank you!

parkr:
@jekyllbot: merge +minor

9bow:
Thank you very much, @parkr!
With your review, this PR was a great experience for me. :)

9bow:
Could you please let me know when the next release schedule is?
I'd love to try out this new feature. 😄

parkr:
v2.14.0 has been released!

9bow:
Awesome! Thanks a lot @parkr!
However, when I try to apply it, a conflict with the [dependency of the github-pages](https://github.com/github/pages-gem/blob/master/lib/github-pages/dependencies.rb#L31) gem.
Could you please help on this part?

9bow:
I've finally rolled out [this cool new feature on our site](https://pytorch.kr/about/contributors)! :tada:
I made it automatically show repositories from @PyTorchKorea, which has 2 or more contributors 
Unfortunately, it's now before the official release, so I'm referring to [my personal repository](https://github.com/PyTorchKorea/pytorch.kr/blob/master/Gemfile#L5).

Again, thank you very much, @parkr!

PR #235, opened 2022-05-05
Add release workflow

parkr:
This allows us to make a release by just updating the file and pushing it to `main`.


parkr:
@jekyllbot: merge +dev

PR #239, opened 2022-05-22
Deprecate unnecessary constant in `GitHubMetadata::Client`

ashmaroli:
I stumbled upon the realization that [`Faraday::ConnectionFailed` is valid ](https://github.com/lostisland/faraday/blob/22b689d78f57b688c82084c8e3ee19e136f2ab5b/lib/faraday/error.rb#L4) even in faraday-0.17 and older and that there is no need of testing for `Faraday::Error::ConnectionFailed`. The latter class is just [an alias](https://github.com/lostisland/faraday/blob/22b689d78f57b688c82084c8e3ee19e136f2ab5b/lib/faraday/error.rb#L60-L63) of the former.

Therefore, let us deprecate the `GitHubMetadata::Client::FARADAY_FAILED_CONNECTION` constant which should have ideally been a private_constant.


parkr:
IIRC, this was for the Faraday 0.x to 1.0 migration.

parkr:
@jekyllbot: merge +fix

Issue #242, opened 2022-10-15
API calls for each and every repository? #2


egor-tensin:
I couldn't find a button to re-open issue https://github.com/jekyll/github-metadata/issues/237, so I'm creating a new one. I finally moved to Jekyll 4, upgraded jekyll-github-metadata to 2.15.0, and was happy for a bit, until I actually started working on my Jekyll projects. When I run `bundle exec jekyll serve --verbose` on repository https://github.com/egor-tensin/egor-tensin.github.io, it makes a single API call, as expected:

```
         Rendering: index.html
  Pre-Render Hooks: index.html
  Rendering Liquid: index.html
   GitHub Metadata: Generating for egor-tensin/egor-tensin.github.io
  Rendering Markup: index.html
Post-Convert Hooks: index.html
  Rendering Layout: index.html
   GitHub Metadata: Calling @client.repository("egor-tensin/egor-tensin.github.io", {:accept=>"application/vnd.github.drax-preview+json"})
           Writing: /home/egor/workspace/personal/egor-tensin.github.io/_site/index.html
                    done in 1.518 seconds.
         Requiring: jekyll-watch
```

However, when I modify something (while Jekyll is serving), it _again_ starts making API calls for every repository (+ extra 7 unneeded API calls):

```
[2022-10-15 22:18:15] DEBUG Jekyll::Commands::Serve::Servlet is invoked.
      Regenerating: 1 file(s) changed at 2022-10-15 22:18:25
                    index.html
   GitHub Metadata: Calling @client.pages("egor-tensin/egor-tensin.github.io", {})
   GitHub Metadata: Calling @client.contributors("egor-tensin/egor-tensin.github.io")
   GitHub Metadata: Calling @client.latest_release("egor-tensin/egor-tensin.github.io")
   GitHub Metadata: Calling @client.organization("egor-tensin")
   GitHub Metadata: Calling @client.organization("egor-tensin")
   GitHub Metadata: Calling @client.user("egor-tensin")
   GitHub Metadata: Calling @client.list_repos("egor-tensin", {:type=>"public", :accept=>"application/vnd.github.mercy-preview+json"})
   GitHub Metadata: Calling @client.releases("egor-tensin/aes-tools")
   GitHub Metadata: Calling @client.contributors("egor-tensin/aes-tools")
   GitHub Metadata: Calling @client.releases("egor-tensin/blog")
   GitHub Metadata: Calling @client.contributors("egor-tensin/blog")
[...]

Issue #237, opened 2022-05-14
API calls for each and every repository?


egor-tensin:
Hey, I'm trying to migrate my stuff from GitHub Pages to an external hosting. Part of that process is getting rid of the github-pages gem and replacing it with a newer Jekyll version. I also use github-metadata. But, when I did the upgrade (in commit https://github.com/egor-tensin/egor-tensin.github.io/commit/d231c9b71d31b4208ef679106ba38a6dbf91ef61), github-metadata started making API calls for each and every repository that I have (this is just an excerpt):

```
 Incremental build: disabled. Enable with --incremental
      Generating...
   GitHub Metadata: Generating for egor-tensin/egor-tensin.github.io
   GitHub Metadata: Calling @client.repository("egor-tensin/egor-tensin.github.io", {:accept=>"application/vnd.github.drax-preview+json"})
   GitHub Metadata: Calling @client.pages("egor-tensin/egor-tensin.github.io", {})
   GitHub Metadata: Calling @client.contributors("egor-tensin/egor-tensin.github.io")
   GitHub Metadata: Calling @client.latest_release("egor-tensin/egor-tensin.github.io")
   GitHub Metadata: Calling @client.organization("egor-tensin")
   GitHub Metadata: Calling @client.organization("egor-tensin")
   GitHub Metadata: Calling @client.user("egor-tensin")
   GitHub Metadata: Calling @client.list_repos("egor-tensin", {:type=>"public", :accept=>"application/vnd.github.mercy-preview+json"})
   GitHub Metadata: Calling @client.releases("egor-tensin/aes-tools")
   GitHub Metadata: Calling @client.contributors("egor-tensin/aes-tools")
   GitHub Metadata: Calling @client.releases("egor-tensin/blog")
   GitHub Metadata: Calling @client.contributors("egor-tensin/blog")
   GitHub Metadata: Calling @client.releases("egor-tensin/build-boost")
   GitHub Metadata: Calling @client.contributors("egor-tensin/build-boost")
```

And it goes on doing 2 calls for every repository of mine, which takes too long. What am I doing wrong?

egor-tensin:
I have the `repository` setting in my _config.yml pointing to the correct repository, it has been that way since forever.

egor-tensin:
As part of the upgrade, my jekyll-github-metadata gem was bumped from version 2.13.0 to 2.14.0. However, if I fix the version at 2.13.0, the problem goes away.

egor-tensin:
[...]

PR #245, opened 2023-01-07
Prevent loading all GH data on reset

dunkmann00:
The way this was setup, when running `build` or `serve` with `--watch`, `inspect` was called and was forcing everything to load. This could cause a very large number of API calls to GitHub. With these changes `GitHubMetadata` will no longer trigger all of those calls on reset.

The other scenario where a bunch of calls were being forced to be made was when putting the `github` key with any sub-keys in `_config.yml`. This went through every key in `GitHubMetadata` and loaded it, once again potentially making a very large number of calls. These changes take care of this scenario as well.

I was running into this problem when trying to make changes to a theme using this. Every time I ran tests on the theme it would incur a bunch of API calls and I would quickly get a rate limit error and have to wait some time for it to go away. I saw this was originally brought up as a problem in #237, there was a fix but it didn't totally fix it so this bug was brought up again in #242. I hope these changes are acceptable because it would be really nice to get rid of this problem.

Let me know if there is anything else I can do to help fix this!


dunkmann00:
Hello! Has anyone had a chance to take a look at this?

@parkr I see you have historically been the one to work on this repo...my apologies in advance if you no longer do!

lib/jekyll-github-metadata/site_github_munger.rb
```diff
@@ -35,14 +35,14 @@ def github_namespace
         when nil
           drop
         when Hash
-          Jekyll::Utils.deep_merge_hashes(drop, site.config["github"])
+          drop.merge(site.config["github"])
```
parkr, line 38: The real fix seems to be this change, since [`#deep_merge_hashes` uses `merge!`](https://github.com/jekyll/jekyll/blob/ff60d51c9b59d3bb5cc89f6a4f493602b42c5cac/lib/jekyll/utils.rb#L321) which resolves all the keys in the `Drop`. Perhaps instead of this, we could modify the `Drop` to fall back to make a copy of the original `site.config["github"]` and use it instead of the computed Drop value. So when a user puts `site.github.repo` in their Liquid template, it queries the Drop, which first looks at `site.config["github"]["repo"]`, then at `Drop#repo` as a fallback. `Drop#repo` would then make the necessary API calls only if called.

lib/jekyll-github-metadata/repository.rb
```diff
@@ -118,7 +118,8 @@ def owner_public_repositories
           :accept => "application/vnd.github.mercy-preview+json",
         }
         memoize_value :@owner_public_repositories, Value.new("owner_public_repositories", proc do |c|
-          c.list_repos(owner, options).map do |r|
+          repos = c.list_repos(owner, options) || []
```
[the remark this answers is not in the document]
dunkmann00: Sorry, yes this change actually was not relevant to this issue. I left it in because it seemed innocent/simple enough. When you don't have a token setup (or hit an API limit) it throws `NoMethodError: undefined method 'map' for false:FalseClass`. This can actually be seen in the AppVeyor CI runs as well: https://ci.appveyor.com/project/jekyll/github-metadata/builds/43615200/job/n59xtdhu8cr1v4fp#L140

If you prefer I could remove this from here and make a separate PR for it?

lib/jekyll-github-metadata/site_github_munger.rb
dunkmann00: Maybe? But that's mostly what is happening with this approach, or so I am pretty sure.

`Drop.merge` uses [merge!](https://github.com/jekyll/jekyll/blob/ff60d51c9b59d3bb5cc89f6a4f493602b42c5cac/lib/jekyll/drops/drop.rb#L250) ultimately as well, so you are getting the same merge behavior, but instead of running it on every key from the drop, it gets run on every key from `site.config["github"]`.

And that merge puts everything in the [mutations](https://github.com/jekyll/jekyll/blob/ff60d51c9b59d3bb5cc89f6a4f493602b42c5cac/lib/jekyll/drops/drop.rb#L154) hash, and [checks that first](https://github.com/jekyll/jekyll/blob/ff60d51c9b59d3bb5cc89f6a4f493602b42c5cac/lib/jekyll/drops/drop.rb#L126) when the drop is queried.

lib/jekyll-github-metadata/metadata_drop.rb
```diff
@@ -10,6 +10,11 @@ class MetadataDrop < Jekyll::Drops::Drop
 
       mutable true
 
+      def initialize(obj, original_config = nil)
+        super(obj)
+        @original_config = original_config
```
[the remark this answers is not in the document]
ashmaroli: Hello @dunkmann00.
What version of this plugin are you using? To my knowledge, the issue with Jekyll Cache calling `config.inspect` was fixed once the plugin stopped depending on `site.config["github"]` from v2.15.0 via https://github.com/jekyll/github-metadata/commit/a771d0d7457a7bff42b129ce0bed70c74514a506.
Therefore, only the Drop being cast into a Hash should trigger the resolution of all drop keys.

*Just asking for clarity.*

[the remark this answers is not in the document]
dunkmann00: Hello @ashmaroli! I am using `2.15.0`. The change via the commit you reference fixes the problem on the initial load and render of the site. However, when the site is rebuilt automatically while running `jekyll serve`, the problem presents itself again.

The original fix waits to inject `site.config["github"]` until after `process` and therefore `reset` is called. But when `jekyll-watch` notices a change and calls `process`, `site.config["github"]` has already been injected. Since it is already injected, we see the original problem occurring once again.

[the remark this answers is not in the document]
parkr:
```ruby
    Jekyll::Hooks.register :site, :pre_render do |_site, payload|
      SiteGitHubMunger.global_munger.inject_metadata!(payload)
    end
```

Is there a post_render hook where we could un-inject this?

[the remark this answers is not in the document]
dunkmann00: 🥳 That seems like a great idea! I think at one point this thought crossed my mind but I didn't understand enough of how the library worked so I didn't pursue it. But looking again this totally seems doable. I'll update the PR with this change.

parkr:
Running the tests. This LGTM otherwise. 

In your tests locally, it prevents the issue?

dunkmann00:
@parkr Yep, when running `jekyll serve --verbose` I can see from the `Github Metadata` log message it is only loading things that are actually used. And this is true when there are `github` variables set in `_config.yml` and also when the page is regenerated after it detects a change.

parkr (commented):
Ok last question – can we add a test for this? It looks like the code for the `--watch` flag calls `site.process` each time: https://github.com/jekyll/jekyll-watch/blob/9c9d2ab3c995784343c7791bc6f02f7c6b2dc1e5/lib/jekyll/watcher.rb#L128

dunkmann00:
Sure I'd be happy to add a test. I'm not sure how to test this though? What did you have in mind?

parkr:
Thanks for jumping through these hoops, a test will be really helpful if we can set it up properly. I'd want to:

1. Run site.process once on the fixture site
2. Assert that each API call happens, and that it happens once (or the proper # of times)
3. Run site.process a second time, asserting that no more API calls occur.

dunkmann00:
I'm not very familiar with ruby testing and I can't find anything similar to what you're asking in other tests to use as a kind of guide. I can try to put something together but I really don't know how to make a test to do this. Would this test go in `site_github_munger_spec.rb`? I think it needs its own context? How do I go about intercepting API calls to count them?

parkr:
Run process twice and use `expect_api_call` time ensure things are called once. https://github.com/jekyll/github-metadata/blob/main/spec/spec_helpers/web_mock_helper.rb 

I'm on mobile now but can take a closer look soon!

dunkmann00:
Okay made some good progress!

I realized that the tests you were describing weren't going to completely test this new functionality because the point of this isn't to make sure API calls only happen once. That was already happening, so that is useful to test but not the whole picture. The point of these changes was to make sure we don't load everything just because `github` was used in `_config.yml` or `reset` is triggered while `jekyll serve` is running.

So in this new commit I:
1. Create a new fixture site that only uses a few `site.github` attributes
2. Process the site and check that those endpoints are loaded once
3. Process the site again and check that those endpoints are still only loaded once
4. Using a new `web_mock_helper.rb` function, `not_expect_api_call`, we check to make sure other endpoints that should **not** have been called are **not** called

Let me know what you think.

lib/jekyll-github-metadata/repository.rb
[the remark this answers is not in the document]
parkr: Let's leave this. Resolving.

spec/site_github_munger_spec.rb
```diff
@@ -231,4 +231,27 @@
...
+
+    it "process site twice (simulate reset), check API calls" do
+      config = Jekyll::Configuration.from({"source" => source, "destination" => dest})
+      site = Jekyll::Site.new(config)
+      site.process
+      expect_api_call "/repos/jekyll/github-metadata"
+      expect_api_call "/repos/jekyll/github-metadata/releases/latest"
+      expect_api_call "/orgs/jekyll"
+      site.process
+      expect_api_call "/repos/jekyll/github-metadata"
```
parkr, line 247: A little more context could be helpful here. Could you add a comment along the lines of "We expect API calls again for the fields we use in the fixture site, but NOT for fields we don't"?

spec/site_github_munger_spec.rb
```diff
@@ -231,4 +231,27 @@
       end.to raise_error(Jekyll::GitHubMetadata::Client::BadCredentialsError)
     end
   end
+
+  context "render the 'uninject' fixture test site" do
+    let(:source) { File.expand_path("test-site-uninject", __dir__) }
+    let(:dest) { File.expand_path("../tmp/test-site-uninject-build", __dir__) }
+
+    it "process site twice (simulate reset), check API calls" do
```
parkr, line 239: Might be worth ensuring that the fields are properly rendered in the resulting HTML after the second `site.process` call?

parkr (approved):
Nicely done, @dunkmann00! I have one comment requesting a comment and another asking your thoughts about ensuring the HTML output is correct after the second `site.process` call. Other than those two things, this LGTM to ship!

dunkmann00:
@parkr I made those changes! I also added another fixture site, this time it is the rendered version of the new fixture site. I used it for checking if the rendered site is correct.

spec/test-site-uninject-rendered/index.html
```diff
@@ -0,0 +1,13 @@
...
...avatars.githubusercontent.com/u/6166343?v=3", "gravatar_id"=>"", "url"=>"https://api.github.com/users/jekyllbot", "html_url"=>"https://github.com/jekyllbot", "followers_url"=>"https://api.github.com/users/jekyllbot/followers", "following_url"=>"https://api.github.com/users/jekyllbot/following{/other_user}", "gists_url"=>"https://api.github.com/users/jekyllbot/gists{/gist_id}", "starred_url"=>"https://api.github.com/users/jekyllbot/starred{/owner}{/repo}", "subscriptions_url"=>"https://api.github.com/users/jekyllbot/subscriptions", "organizations_url"=>"https://api.github.com/users/jekyllbot/orgs", "repos_url"=>"https://api.github.com/users/jekyllbot/repos", "events_url"=>"https://api.github.com/users/jekyllbot/events{/privacy}", "received_events_url"=>"https://api.github.com/users/jekyllbot/received_events", "type"=>"User", "site_admin"=>false}, "prerelease"=>false, "created_at"=>2017-01-18 20:10:29 UTC, "published_at"=>2017-01-18 20:10:33 UTC, "assets"=>[], "tarball_url"=>"https://api.github.com/repos/jekyll/github-metadata/tarball/v2.3.1", "zipball_url"=>"https://api.github.com/repos/jekyll/github-metadata/zipball/v2.3.1", "body"=>"- Remove log on Octokit::NotFound (#86)\n"}</h3>
```
ashmaroli, line 11: Sorry @dunkmann00, but I will have to block this change from getting merged.
This `"Latest release"` metadata looks like it will break when a new release ships.

ashmaroli (changes requested):

spec/test-site-uninject-rendered/index.html
parkr: I think this is fixture data, `site.github.latest_release`, which is returned from https://github.com/jekyll/github-metadata/blob/main/spec/webmock/api_get_latest_release.json. This should be stable since we're stubbing webmock with this fixture file. That's further supported by the fact that the latest release is 2.15.0: https://github.com/jekyll/github-metadata/releases (it should be reporting 2.3.1 if we were using "live" data).

ashmaroli: Oh! Didn't realize this was fixture JSON.
Removing the blocking review now..

parkr:
@jekyllbot: merge +minor

(I'm going minor here as a reminder to bump the minor version. It's a bug fix in a way but it's also a slight change in behavior that folks might find surprising.)

dunkmann00:
Awesome! Thanks @parkr and @ashmaroli for helping with and reviewing this! And for all that you do for Jekyll ❤️

parkr:
Thanks for working through this with us!

PR #252, opened 2023-02-28
Bump rubocop-jekyll to 0.13.0 and fix rubocop issues

dunkmann00:
I'm opening this PR in the hopes of helping move along a new release of `github-metadata`.

This has the same version bump of `rubocop-jekyll` as #251 but I also went ahead and let rubocop autocorrect the formatting issues it found.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,9 @@ AllCops:
     - vendor/**/*
     - script/*
 
+Style/FetchEnvVar:
+  Enabled: false
+
 Metrics/LineLength:
   Enabled: false
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,10 +10,7 @@ AllCops:
     - vendor/**/*
     - script/*
 
-Style/FetchEnvVar:
-  Enabled: false
-
-Metrics/LineLength:
+Layout/LineLength:
   Enabled: false
 
 Metrics/BlockLength:
@@ -25,6 +22,9 @@ Metrics/MethodLength:
     - spec/**/*
 
 Style/AccessModifierDeclarations:
+  Enabled: false
+
+Style/FetchEnvVar:
   Enabled: false
 
 Style/FrozenStringLiteralComment:
--- a/jekyll-github-metadata.gemspec
+++ b/jekyll-github-metadata.gemspec
@@ -24,5 +24,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "pry"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "rspec", "~> 3.12.0"
-  spec.add_development_dependency "rubocop-jekyll", "~> 0.12.0"
+  spec.add_development_dependency "rubocop-jekyll", "~> 0.13.0"
 end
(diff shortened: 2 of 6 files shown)


dunkmann00:
Just curious, why is appveyor used for CI? It seems like it has something to do with testing for Windows, but GitHub Actions builds for Windows also?

ashmaroli:
Hello @dunkmann00,

Thank you for submitting this PR. However, I am out of context here. That means the title + description of this pull request is inadequate. Anyone stumbling across this pull request (including maintainers, at a future date) is going to scratch their heads wondering about the **purpose / aim** of this PR. (No, *assuming* context from included commit messages isn't the right way.)

On the contrary to suggested fixes from RuboCop, resorting to `ENV.fetch` isn't ideal here because the changed lines here all default to `nil` which is already the implicit result of these lines prior to being changed. If you were to follow the link(s) given in the CI report for PR #251, you shall understand the reason why RuboCop chose to suggest change(s). Regardless, I'd like you to revert use of `ENV.fetch` and explicitly disable the cop in `.rubocop.yml`. That said, let us benchmark the two situations for clarity first.

The reasons why AppVeyor is still around when GH Actions exist:
- Notice how AppVeyor has just one commit check status as against the many for GHA jobs? Notice how it's now slightly inconvenient  when wishing to access reports for **just Ubuntu-based tests** (Hint: visual noise).
- Perhaps, the above point is just my problem, i.e. subjective. But, let us actually compare two  *passing test reports for a given Ruby version*, say Ruby 2.7. Notice (expand step `Run script/cibuild`) how while the Ubuntu-based report reads as expected, the Windows-based test hasn't actually run the underlying test-suite?!
- Notice how the passing check for Windows-based tests led to the false assumption that the GH Action workflow was set up correctly and has run without errors, whereas AppVeyor has attempted to run the test-suite but failed due to version mismatch in `appveyor.yml`.

dunkmann00:
Hello @ashmaroli! My apologies on the title. I had used 'Checkout with GitHub Desktop' to get dependabot's commit and in the process of doing that I never really gave the title any attention (the current title was just auto generated). I'll fix that.

I did look at Rubocop's explanation, the change to `ENV.fetch` seemed unnecessary but I figured I'd let it make the auto change and then go back and correct anything as requested.

> Notice (expand step Run script/cibuild) how while the Ubuntu-based report reads as expected, the Windows-based test hasn't actually run the underlying test-suite?!

😆 I didn't realize it didn't actually test anything! Well that makes perfect sense then. Would you like me to change anything in `appveyor.yml` to run Ruby 2.7 instead 2.6?

ashmaroli:
> Would you like me to change anything in `appveyor.yml` to run Ruby 2.7 instead 2.6?

@dunkmann00 Actually, getting AppVeyor to run Ruby 2.7 is a bit tricky. I'll handle it via a separate PR and will ping you after its merged.

ashmaroli:
@dunkmann00 AppVeyor has been reconfigured on the `main` branch. You may incorporate those changes into the branch here by running the following in your local terminal:

```bash
git pull upstream main   # 'upstream' refers to the base repo in Jekyll org. 
```

dunkmann00:
@ashmaroli Okay I incorporated your changes and updated the other rubocop issues. Let me know if you are okay with the current changes or if you would like any of them reverted.

ashmaroli:
Current changes look fine to me. However, there's a warning from RuboCop about incorrect cop namespace in `.rubocop.yml`. The current PR is the ideal place to fix the cop name. Please rename and move directive to appropriate location in `.rubocop.yml` (alphabetical order)

dunkmann00:
Should be all set now.

ashmaroli (approved):

ashmaroli:
Thanks @dunkmann00
@jekyllbot: merge +dev

dunkmann00:
Thanks @ashmaroli! Any idea when you'll be able to release a new version?

PR #253, opened 2023-03-01
Bump Ruby versions in AppVeyor jobs

ashmaroli:
Ruby 2.7 is nearing EOL (in 30 days).
In spite of that drop only older rubies from AppVeyor CI config since Ruby 2.7 is the last series of Ruby 2.


ashmaroli:
@jekyllbot: merge +dev

PR #260, opened 2023-06-30
Update references of help.github.com to docs.github.com

JamesMGreene:
[Back in 2020](https://github.blog/2020-07-02-how-we-launched-docs-github-com/), the GitHub documentation site formerly hosted at `help.github.com` moved to `docs.github.com`.

While the links that are being generated by this plugin will still correctly redirect to that new site, those redirects may not be maintained forever. Furthermore, hitting `help.github.com` directly will actually redirect to a _different_ site at `support.github.com` instead of the documentation site, which could be confusing. 🤷🏻 

This PR updates all references of `help.github.com` to `docs.github.com`. 🛠️


mattr- (approved):
Thanks for updating these! :heart:

@jekyllbot: merge +fix

Issue #241, opened 2022-08-23
Flag for disabling internet_connected


peterchenadded:
The docs suggest there is support for github enterprise.

But for people using github enterprise they are unlikely want to connect to the internet.

See https://github.com/jekyll/github-metadata/blob/f3b27ce021158c1206cb1641f6734fe4f09f59f9/lib/jekyll-github-metadata/client.rb#L104

Is it possible to disable the internet_connected check?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code>/<code>main</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

Issue #127, opened 2018-05-25
Disable network requests


pathawks:
For purposes of benchmarking, it would be nice if the time it takes to request data from the GitHub API did not contribute to the build time. Could an option be added to explicitly set `@internet_connected` to `False`?

https://github.com/jekyll/github-metadata/blob/47855c340dc4ac4e7c7b4db7ab6349c0286ca50e/lib/jekyll-github-metadata/client.rb#L96-L97

Could there be a way to build the site once with networking, retrieve the requested data, and then use only cached data for subsequent builds?

benbalter:
@pathawks glad to review a PR to make that easier to set.... alternatively, VCR or Webmock may help spoof the requests to benchmark the underlying logic without the network request.

pathawks:
@benbalter Could you provide just a bit more detail for how that would work?

For the record, my use case is benchmarking Jekyll sites that depend on this plugin, rather than benchmarking the plugin itself. I want to be able to run something like `jekyll build` and not worry that an inconsistent network connection could be skewing results.

benbalter:
@pathawks are you running it from command line or within a ruby script? If within a ruby script (e.g., bench-marking the actual site execution) prior to starting the timer, you'd require Webmock and disable network connections, stubbing in those you'd expect with JSON files locally. If timing via, e.g., shell, including the Ruby boot time, it wouldn't work, and the option you described above would be the best route.

pathawks:
Plan was from Bash script, but I may need to reconsider.

Closing this until I know if any specific changes need to be made here.

PR #285, opened 2025-02-21
Allow disabling of all network accesses

ibbem:
## Context
We use [Nix](https://nixos.org/) to build our [project](https://github.com/VariantSync/DiffDetective) reproducibly, including the github-pages documentation. [This setup](https://github.com/VariantSync/DiffDetective/blob/07756a1b6c3b89049ba04d4980a220c499fc7f36/default.nix#L96) is also used in our [CI](https://github.com/VariantSync/DiffDetective/blob/07756a1b6c3b89049ba04d4980a220c499fc7f36/.github/workflows/maven.yml#L26) to build the github-pages which are eventually deployed. Nix prohibits network connections (except if the output hash is specified, which is unpractical in this case) because the network is not reproducible. However, in some circumstances (e.g., the default installation on macOS) the Nix sandbox is disabled or weakened. Hence, github-metadata accesses the network which results in failing builds or differing outputs.

## Feature Request
We would like to reliably disable network accesses without actually disabling the network.

## Implemented solution
Network accesses are skipped if the environment variable `PAGES_DISABLE_NETWORK` is set.

## Related issues
- Same feature request, although different motivation: #241
- Different use case, probably solved differently: #127

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -1,3 +1,9 @@
+## HEAD
+
+## Minor Enhancements
+
+  * Disable network related features if the env var `PAGES_DISABLE_NETWORK` is set
+
 ## 2.16.1 / 2023-12-22
 
 ### Bug Fixes
--- a/History.markdown
+++ b/History.markdown
@@ -1,9 +1,3 @@
-## HEAD
-
-## Minor Enhancements
-
-  * Disable network related features if the env var `PAGES_DISABLE_NETWORK` is set
-
 ## 2.16.1 / 2023-12-22
 
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


ashmaroli:
Hello @ibbem, please revert the change to `History.markdown`. It gets updated automatically with the pull-request-title when the pull request is merged. Additionally, this requires a test to verify that setting the env var does what is intended.

P.S. You need not rebase your branch. All commits will be squashed automatically at merge.

ashmaroli:
No @ibbem, this LGTM!
Thank you for your contribution.

@jekyllbot: merge +minor

